PR #77, opened 2014-12-03
Add "booting" flag and emit "booted" event

superkhau:
See strongloop/loopback-workspace#170

### DIFF
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -52,6 +52,32 @@ describe('executor', function() {
     }
   });
 
+  describe('when booting', function() {
+    it('should set the booting status', function(done) {
+      expect(app.booting).to.be.undefined();
+      boot.execute(app, dummyInstructions, function(err) {
+        expect(err).to.be.undefined();
+        expect(app.booting).to.be.false();
+        done();
+      });
+    });
+
+    it('should emit the `booted` event', function(done) {
+      app.on('booted', function() {
+        // This test fails with a timeout when the `booted` event has not been
+        // emitted correctly
+        done();
+      });
+      boot.execute(app, dummyInstructions, function(err) {
+        expect(err).to.be.undefined();
+      });
+    });
+
+    it('should work when called synchronously', function() {
+      boot.execute(app, dummyInstructions);
+    });
+  });
+
   it('configures models', function() {
     boot.execute(app, dummyInstructions);
     assert(app.models);
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -16,6 +16,10 @@ var path = require('path');
  */
 
 module.exports = function execute(app, instructions, callback) {
+  callback = callback || function() {};
+
+  app.booting = true;
+
   patchAppLoopback(app);
   assertLoopBackVersion(app);
 
@@ -37,7 +41,15 @@ module.exports = function execute(app, instructions, callback) {
     function(done) {
       enableAnonymousSwagger(app, instructions);
       done();
-    }], callback);
+    }], function(err) {
+      app.booting = false;
+
+      if (err) return callback(err);
+
+      app.emit('booted');
+
+      callback();
+    });
 };
 
 function patchAppLoopback(app) {


superkhau:
@bajtos Added unit test, reassigning to you for review.

test/executor.test.js
```diff
@@ -52,6 +52,39 @@ describe('executor', function() {
     }
   });
 
+  describe('when booting', function() {
+    it('should set the booting status', function(done) {
+      expect(app.booting).to.be.undefined;
```
bajtos: Please use `.to.be.undefined()` to prevent jshint warnings.

```
$ npm test

> loopback-boot@2.5.2 pretest /Users/bajtos/src/loopback/boot
> jscs . && jshint .

No code style errors found.
test/executor.test.js: line 57, col 33, Expected an assignment or function call and instead saw an expression.
test/executor.test.js: line 59, col 27, Expected an assignment or function call and instead saw an expression.
test/executor.test.js: line 60, col 35, Expected an assignment or function call and instead saw an expression.
test/executor.test.js: line 67, col 28, Expected an assignment or function call and instead saw an expression.
test/executor.test.js: line 71, col 27, Expected an assignment or function call and instead saw an expression.
test/executor.test.js: line 77, col 27, Expected an assignment or function call and instead saw an expression.

6 errors
npm ERR! Test failed.  See above for more details.
```

test/executor.test.js
```diff
@@ -52,6 +52,39 @@ describe('executor', function() {
...
+      boot.execute(app, dummyInstructions, function(err) {
+        expect(err).to.be.undefined;
+        expect(app.booting).to.be.false;
```
bajtos: `.to.be.undefined()`, `.to.be.false()`.

test/executor.test.js
```diff
@@ -52,6 +52,39 @@ describe('executor', function() {
...
+        expect(err).to.be.undefined;
+      });
+    });
+
+    it('should work when called asynchronously', function(done) {
+      boot.execute(app, dummyInstructions, function(err) {
+        expect(err).to.be.undefined;
+        done();
+      });
+    });
```
bajtos: What is the purpose of this test? Isn't it just duplicating the existing test "executor with boot with callback should run `boot/*` files asynchronously"?

bajtos:
see my comments above.

test/executor.test.js
superkhau: Didn't see this because I ran `mocha` instead. Fixed.

test/executor.test.js
superkhau: Fixed.

test/executor.test.js
superkhau: Yes, it's doing the same thing. Removing.

superkhau:
@bajtos Review again please. https://github.com/strongloop/loopback-boot/commit/196d19276f334ba001ef90e6dc19643d4d272b99

test/executor.test.js
```diff
@@ -52,6 +52,31 @@ describe('executor', function() {
...
+      boot.execute(app, dummyInstructions, function(err) {
+        expect(err).to.be.undefined();
+        expect(app.booting).to.be.false();
+        done();
+      });
+    });
+
+    it('should emit the `booted` event', function(done) {
+      app.on('booted', function() {
+        expect(true).to.be.ok();
```
bajtos: This assertion never fails, please remove.

If you feel a test without an assertion is confusing, then add a comment at L67 explaining that the test fails on timeout when the event is not emitted.

bajtos:
There is no test checking that `app.booting` is actually set to `true`, but I suppose I can live with that.

Please fix the comment on L67 above and squash all commits into a single one. Other than that, the patch LGTM.

test/executor.test.js
superkhau: Done.

superkhau:
I wanted to add a test for `app.booting = true`, but I had no idea how to test it in the middle of the function execution. I'm all ears if you have a good technique to do that.

superkhau:
@bajtos Updated with latest comments. Please review again before merge.

bajtos:
Landed, thank you. I'll make a new npmjs release a bit later, once I have all pieces related to #86 in place.

bajtos:
> I wanted to add a test for `app.booting = true`, but I had no idea how to test it in the middle of the function execution. I'm all ears if you have a good technique to do that.

Add a boot script asserting on the flag.

```js
// boot/set-app-is-booting.js
module.exports = function(app) {
  app.wasBootingInBootScript = app.booting;
}
```

Then run `boot(app, appdir)` and check that `app.wasBootingInBootScript` was set to true.

Issue #45, opened 2014-09-29
Component Loading


ritch:
`boot()` should load dependent components, make their models available for configuration, [mount middleware](https://github.com/strongloop/loopback-boot/issues/44), etc.

[See here for details](https://github.com/strongloop/loopback-workspace/issues/149).

Depends on https://github.com/strongloop/loopback-example-component/issues/10

PR #86, opened 2015-01-06
Configure components via `component-config.json`

### LINKED ISSUE #81: Inject `app` object to middleware factories
Per @raymondfeng comment in #75:
> I would also need to access the app instance either via the parameter injection or set by boot to the factory/handler instance.

bajtos:
Load configuration of components from `component-config`
and configure all components as specified.

Sample JSON:

    {
      "loopback-component-foobar": {
        "option1": "value1",
        "option2": "value2"
      }
    }

The component is expected to export the following function:

    module.exports = function(app, options) { /* ... */ };

Close #81, see also #45.

/to @raymondfeng @ritch please review

### DIFF
--- a/index.js
+++ b/index.js
@@ -109,6 +109,8 @@ var addInstructionsToBrowserify = require('./lib/bundler');
  * for files containing model definitions.
  * @property {Object} [middleware] Middleware configuration to use instead
  * of `{appRootDir}/middleware.json`
+ * @property {Object} [components] Component configuration to use instead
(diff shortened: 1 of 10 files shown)


bajtos:
Note: I have deliberately chosen the longer form `component-config.json` over `components.json`, as the latter could be confused with `component.json` used by http://componentjs.com/

raymondfeng:
LGTM

PR #88, opened 2015-01-08
Don't swallow error when a sub-dependency doesn't resolve.

STRML:
This prevents an occurence where an error is completely swallowed if a script
required by loopback-boot has a bad require() call. The script is never ran
but execution continues.

I had a situation where a typo caused a model file to be unloaded, and there was
absolutely no indication this was happening until I tried to call a method via the REST 
adapter and got a message that the sharedClass had no method by that name.

This patch only swallows the error if the actual module we're trying to require
is missing, it does not swallow if a sub-require fails.

It would be nice to merge this into the 1.x branch as well.

### DIFF
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -175,6 +175,17 @@ describe('executor', function() {
     expect(app.models.Customer._modelsWhenAttached).to.include('UniqueName');
   });
 
+  it('throws on bad require() call inside boot script', function() {
+    var file = appdir.writeFileSync('boot/badScript.js',
+      'require("doesnt-exist"); module.exports = {};');
+
+    function doBoot() {
+      boot.execute(app, someInstructions({ files: { boot: [file] } }));
+    }
+
+    expect(doBoot).to.throw(/Cannot find module \'doesnt-exist\'/);
+  });
+
   it('instantiates data sources', function() {
     boot.execute(app, dummyInstructions);
     assert(app.dataSources);
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -224,7 +224,7 @@ function runScripts(app, list, callback) {
   var functions = [];
   list.forEach(function(filepath) {
     debug('Requiring script %s', filepath);
-    var exports = tryRequire(filepath);
+    var exports = require(filepath);
     if (typeof exports === 'function') {
       debug('Exported function detected %s', filepath);
       functions.push({
@@ -249,19 +249,6 @@ function runScripts(app, list, callback) {
       done();
     }
   }, callback);
-}
-
-function tryRequire(modulePath) {
-  try {
-    return require.apply(this, arguments);
-  } catch (e) {
-    if (e.code === 'MODULE_NOT_FOUND') {
-      debug('Warning: cannot require %s - module not found.', modulePath);
-      return undefined;
-    }
-    console.error('failed to require "%s"', modulePath);
-    throw e;
-  }
 }
 
 function setupMiddleware(app, instructions) {


bajtos:
I am wondering whether it makes sense to keep the condition at all? The compiler is pretty good at filtering out files that don't exist. When the file gets all the way to the executor, it is very likely that it was intended to be loaded and therefore it's better to be verbose about the failure.

> It would be nice to merge this into the 1.x branch as well.

:+1:

STRML:
I was wondering about that as well. At least in my 1.x app, this code path is never ran.

bajtos:
Can you please rework the patch to remove the whole branch then?

STRML:
No problem.

STRML:
Do you see any reason to use `tryRequire` at all anymore, then, if we're just going to throw? The module path is available in the require error anyway.

STRML:
I've removed `tryRequire` entirely. If it looks good to you, I'll rebase this into a single commit so you can merge it.

bajtos:
I'll take a look tomorrow.

bajtos:
Is there a way how to write a unit-test to verify this new behaviour? i.e. that `boot` and/or `executor` throw when a file (e.g. a boot script) is missing?

STRML:
@bajtos Just added one.

bajtos:
Your tests passes even without the change. `tryRequire` is currently used for boot scripts. Which brings us to another issue - when the boot script cannot be required, the error should be sent to `callback`.

bajtos:
@STRML I am afraid your test passes even without the change, thus you are testing a wrong thing. AFAIK `tryRequire` is currently used for boot scripts, not for model scripts. 

Also when the boot script cannot be required, the error should be preferably sent to `callback`, not thrown. That way it is possible to handle boot errors even in case when the boot scripts are running asynchronously in a later tick.

STRML:
I see what the issue is.

I use the 1.x branch today so I was seeing the problem on any kind of boot script (models, datasources, etc.). In the 2.x/master branch, the issue only manifests on explicit `boot` scripts, all other scripts go down a separate code path where they just use `require`.

I've fixed the test in master and I'm also opening another PR to merge this fix into the 1.x branch, where it is a much bigger problem.

As for the error being sent to callback - the only way to do that is to surround every function that calls `require` in a try/catch. There's also the possibility that the error could get ignored in the callback and this whole problem would manifest again. IMO I think it is much better to just halt. A bad `require()` is a code mistake and not something that can just happen unexpectedly at runtime.

bajtos:
> I've fixed the test in master and I'm also opening another PR to merge this fix into the 1.x branch, where it is a much bigger problem.

Excellent, I'll review both PRs.

> As for the error being sent to callback - the only way to do that is to surround every function that calls `require` in a try/catch. There's also the possibility that the error could get ignored in the callback and this whole problem would manifest again. IMO I think it is much better to just halt. A bad `require()` is a code mistake and not something that can just happen unexpectedly at runtime.

Makes sense. Let's do it the way you are proposing, we can always add try/catch statements later, when there is a better reason for that.

test/executor.test.js
```diff
@@ -175,6 +175,21 @@ describe('executor', function() {
     expect(app.models.Customer._modelsWhenAttached).to.include('UniqueName');
   });
 
+  it('throws on bad require() call inside boot script', function() {
+    appdir.writeFileSync('models/BadCustomer.js',
```
bajtos: Please follow the convention of keeping boot scripts in the `boot/` directory (not in `models/`).

STRML: Fixed

bajtos:
:+1: 

Could you please squash all commits into a single one?

STRML:
Done.

bajtos:
Landed and released in `loopback-boot@2.6.2`. Thank you!

STRML:
Thanks!

PR #89, opened 2015-01-12
Fix "incompatible loopback version" check & msg

bajtos:
Use `util.format` to build the error message, `Error` constructors
does not support placeholders like `%s`.

Detect pre-release versions and handle them in the same way as regular
releases.

/to @ritch or @raymondfeng please review


raymondfeng:
LGTM

PR #90, opened 2015-01-13
Backport PR #88 to 1.x branch.

STRML:
This fix is more critical for the 1.x branch than the master branch, as `tryRequire` was surrounding all scripts, not just explicit `boot` scripts. There is a lot more potential for this bug to bite a user in 1.x.

See #88

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -1,7 +1,6 @@
 var assert = require('assert');
 var _ = require('underscore');
 var semver = require('semver');
-var debug = require('debug')('loopback:boot:executor');
 
 /**
  * Execute bootstrap instructions gathered by `boot.compile`.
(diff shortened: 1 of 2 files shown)


bajtos:
Landed, thank you.

STRML:
Thanks a lot.

STRML:
If you get a chance, could you do a release on the 1.x branch?

bajtos:
I did a mistake by releasing 2.x first. Now if I release 1.x, it will replace the readme on npmjs.org with the old version :(

I can't think of any better solution, so I'll release 1.x and then immediately 2.x to get a correct readme.

bajtos:
`loopback-boot@1.1.2`

Issue #49, opened 2014-10-03
Extend options arg to support custom model definitions


bajtos:
Extend the `options` argument of `boot`/`compile` to allow the developer to supply custom model definitions (to replace the result of `findModelDefinitions`). Together with the existing `options.models`, it will allow you to specify all model-related data and still use the benefits which loopback-boot provides (most notably loading models in order defined by inheritance, i.e. base models are loaded before models that are extending them).

```js
boot(app, {
  models: {
    Car: { dataSource: 'db' },
    // etc.
  },
  modelDefinitions: [
     {
       definition: {
         name: 'Car',
         properties: { /*...*/ }
       },
       sourceFile: 'path/to/car.js'
    },
    // etc.
  ]
});
```

This is a follow-up for #27.

shlomiassaf:
@bajtos,

I sent PR for this enhancement, #96 
Could you please check?

Thanks.

PR #96, opened 2015-02-18
Extend options arg to support custom model definitions

shlomiassaf:
* Added  the ability to create ad-hoc modelDefinitions.
* Added the ability to "extend" models-config (choose overwrite/extend)

See #49

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -41,8 +41,8 @@ module.exports = function compile(options) {
 
   if (options.models && options.models['__extend'] === true){
     modelsConfig = _.extend(ConfigLoader.loadModels(modelsRootDir, env), modelsConfig);
-    delete modelsConfig["__extend"];
   }
+  delete modelsConfig["__extend"];
   assertIsValidModelConfig(modelsConfig);
 
   var dsRootDir = options.dsRootDir || appRootDir;
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -38,6 +38,11 @@ module.exports = function compile(options) {
   var modelsRootDir = options.modelsRootDir || appRootDir;
   var modelsConfig = options.models ||
     ConfigLoader.loadModels(modelsRootDir, env);
+
+  if (options.models && options.models['__extend'] === true){
+    modelsConfig = _.extend(ConfigLoader.loadModels(modelsRootDir, env), modelsConfig);
+    delete modelsConfig["__extend"];
+  }
   assertIsValidModelConfig(modelsConfig);
 
   var dsRootDir = options.dsRootDir || appRootDir;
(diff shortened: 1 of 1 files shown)


slnode:

shlomiassaf:
Example: 
```
boot(app, {
  appRootDir: __dirname,
  models: {
    "__extend": true,
    "testadhoc": {
      "dataSource": "db",
      "public": true
    },
    "testAdHocManual": {
      "dataSource": "db",
      "public": true
    }
  },
  modelDefinitions: {
    sourceDir: "../common/models",
    definitions: [
      {
        definition: {
          "name": "testadhoc",
          "plural": "testAdhocs",
          "base": "PersistedModel",
          "idInjection": true,
          "properties": {
            "test3": {
              "type": "string"
            },
            "test4": {
              "type": "number"
            }
          },
          "validations": [],
          "relations": {},
          "acls": [],
          "methods": []
        }
      },
      {
        definition: {
          "name": "testAdHocManual",
          "plural": "testAdHocManuals",
          "base": "PersistedModel",
          "idInjection": true,
          "properties": {
            "test5": {
              "type": "string"
            },
            "test6": {
              "type": "number"
            }
          },
          "validations": [],
          "relations": {},
          "acls": [],
          "methods": []
        },
        sourceFile: "../testAdHocManual123.js"
      }
    ]
  }
});

```

modelDefinitions.sourceDir is relative to rootDir
modelDefinitions.definition[x].sourceFile is relative to modelDefinitions.sourceDir || rootDir

when setting `models["__extend"]` to `true`, models are a distinct union of models-config.js and the supplied model config collection object.

bajtos:
@shlomiassaf thank you for the pull request. It is very unclear to me what problem (use-case) are you trying to solve? My initial impression is that you are making things too complex and thus I am inclining to reject this patch.

shlomiassaf:
@bajtos,

This is an issue/enhancement you opened, as stated in my 1st comment: SEE #49 

The use case is very clear, allow loading models using the model definition schema but not from a file, from an object.
It enables loading models from database, remote calls, etc...

bajtos:
I see, thanks for the explanation. I'll review the patch with that use case in mind.

bajtos:
@slnode test please

bajtos:
This patch seems to either implement a different thing, or else it is adding much more features beyond simple "custom model definitions".

Let's split the work into multiple small incremental pull requests, so that I can review each new feature in isolation.

Please start with the feature "allow the developer to supply custom model definitions to replace the result of findModelDefinitions" and add a unit-test to demonstrate & verify the new functionality.

shlomiassaf:
I understand what you mean.

A quick question,
Does replacing the result of `findModelDefinitions` means replacing internal model definitions (`User`, `ACL`, etc..)?

The code as is now, does not allow loading `models` along with built in models since the models-config is either from file or in `options` object.

If a user wants to generate models from database, remote call or any other external source, he will be able to do it but he will have to give up the built-in models.

The solution in my PR enables loading them without skipping the `models-config.json` load.

bajtos:
> A quick question,
Does replacing the result of findModelDefinitions means giving up on internal model definitions (User, ACL, etc..)?

Yes.

> The code as is now, does not allow loading models along with built in models since the models-config is either from file or in options object.

> If a user wants to generate models from database, remote call or any other external source, he will be able to do it but he will have to give up the built-in models.

The built-in models are always defined created by loopback, you should not include them in model definitions passed to loopback-boot. Just reference them in `model-config.json`.

Even if that was not the case, the definition of built-in models can be easily reconstructed:

```js
{
  User: {
    definition: require('loopback/common/models/user.json'),
    sourceFile: require.resolve('loopback/common/models/user.js')
  }
}
```

PR #97, opened 2015-02-19
Add unit test to check `app.booting` flag

superkhau:
Connected to strongloop/loopback-boot#77

### DIFF
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -53,11 +53,12 @@ describe('executor', function() {
   });
 
   describe('when booting', function() {
-    it('should set the booting status', function(done) {
+    it('should set the `booting` flag during execution', function(done) {
       expect(app.booting).to.be.undefined();
-      boot.execute(app, dummyInstructions, function(err) {
+      boot.execute(app, simpleAppInstructions(), function(err) {
         expect(err).to.be.undefined();
-        expect(app.booting).to.be.false();
+        expect(process.bootingFlagSet).to.be.ok();
+        expect(app.booting).to.not.be.ok();
         done();
       });
     });
--- a/test/fixtures/simple-app/boot/booting.js
+++ b/test/fixtures/simple-app/boot/booting.js
@@ -1,0 +1,6 @@
+module.exports = function(app, cb) {
+  if (app.booting)
+    process.bootingFlagSet = true;
+
+  process.nextTick(cb);
+};


superkhau:
@bajtos Review please.

test/executor.test.js
```diff
@@ -53,11 +53,17 @@ describe('executor', function() {
...
-    it('should set the booting status', function(done) {
+    it('should set the `booting` flag during execution', function(done) {
       expect(app.booting).to.be.undefined();
-      boot.execute(app, dummyInstructions, function(err) {
+
+      boot.execute(app, simpleAppInstructions(), function(err) {
         expect(err).to.be.undefined();
+
+        var bootingFlagSet = process.bootFlags.indexOf('booting') > -1;
+        expect(bootingFlagSet).to.be.ok();
```
bajtos: `expect(process.bootFlags).to.include('booting')`

bajtos: Please consider learning yourself more familiar with the different assertions provided by Chai, see http://chaijs.com/api/bdd/. The benefit of these assertions (compared to your solution above) is a much more helpful error message reported when the check fails.

superkhau: Will do. ;) Thanks for the suggestion.

superkhau:
@bajtos Updated based on code review comments. Good to merge?

superkhau:
@slnode test please

bajtos:
There's one more thing I don't like about the current approach, sorry for realising it too late. By reusing `bootFlags`, you now have to change other unrelated tests, plus these other tests are now tightly coupled with the order in which the newly added boot script is executed. This will make the maintenance more difficult.

Please rework the test to use different (new) flag, so that you don't have to modify any existing tests in this PR.

bajtos:
And also please rebase on top of the current master to pull in the fix for the browserify-related test failures.

superkhau:
@bajtos Reworked. Review again please.

superkhau:
@slnode test please

test/executor.test.js
```diff
@@ -53,11 +53,12 @@ describe('executor', function() {
...
   describe('when booting', function() {
-    it('should set the booting status', function(done) {
+    it('should set the `booting` flag during execution', function(done) {
       expect(app.booting).to.be.undefined();
-      boot.execute(app, dummyInstructions, function(err) {
+      boot.execute(app, simpleAppInstructions(), function(err) {
         expect(err).to.be.undefined();
-        expect(app.booting).to.be.false();
+        expect(process.bootingFlagSet).to.be.ok();
+        expect(app.booting).to.not.be.ok();
```
bajtos: I am curious, why did you have to change `.to.be.false()` to `.to.not.be.ok()`?

bajtos:
LGTM.

test/executor.test.js
superkhau: That was by accident actually. I looked up `ok()` and `...not.be.ok()` and `true()`/`false()` makes more sense but I forgot to change it back. I will submit another PR to change it back before we forget about it.

PR #100, opened 2015-02-24
Extend options arg to support custom model definitions

shlomiassaf:
Extend the options argument of boot/compile to allow the developer to supply custom model definitions (to replace the result of findModelDefinitions). Together with the existing options.models, it will allow you to specify all model-related data and still use the benefits which loopback-boot provides (most notably loading models in order defined by inheritance, i.e. base models are loaded before models that are extending them).

See @bajtos issue, #49

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -265,6 +265,24 @@ function sortByInheritance(instructions) {
     });
 }
 
+/**
+ * Returns matching files with a supported `sourceFile` extension like `.js` or `.coffee`
+ * @param allFiles {Array}
+ * @param basename {string}
+ * @returns {Array}
+ */
+function filterValidSourceFiles(allFiles, basename) {
+  var base, ext, validFileType;
+  return allFiles
+    .filter(function(f) {
+      ext = path.extname(f);
+      base = path.basename(f, ext);
+      validFileType = (ext !== '.node') && (ext !== '.json') &&
+      ((typeof require.extensions[ext]) === 'function');
+      return validFileType && (base === basename);
+    });
+}
+
 function verifyModelDefinitions(rootDir, modelDefinitions) {
   if (!modelDefinitions || modelDefinitions.length < 1) {
     return undefined;
@@ -274,7 +292,13 @@ function verifyModelDefinitions(rootDir, modelDefinitions) {
   modelDefinitions.forEach(function(definition, idx) {
     try {
       var fullPath = path.resolve(rootDir, definition.sourceFile);
-      definition.sourceFile = require.resolve(fullPath);
+      var basename = path.basename(fullPath, path.extname(fullPath));
+      var files = tryReadDir(path.dirname(fullPath));
+      var sourceFile = filterValidSourceFiles(files, basename)[0];
+
+      definition.sourceFile = path.join(path.dirname(fullPath), sourceFile);
+      definition.sourceFile = require.resolve(definition.sourceFile);
+
     } catch (err) {
       debug('Model source code not found: %s - %s', definition.sourceFile, err.code || err);
       definition.sourceFile = undefined;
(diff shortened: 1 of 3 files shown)


slnode:

bajtos:
@slnode test please

lib/compiler.js
```diff
@@ -263,6 +265,36 @@ function sortByInheritance(instructions) {
...
+
+  var registry = {};
+  modelDefinitions.forEach(function(definition, idx) {
+    try {
+      var fullPath = path.resolve(rootDir, definition.sourceFile);
+      definition.sourceFile = require.resolve(fullPath);
+    } catch (err) {
+      debug('Model source code not found: %s - %s', definition.sourceFile, err.code || err);
+      definition.sourceFile = undefined;
+    }
```
bajtos: This block should be run only when `definition.sourceFile` is defined. Please add a unit-test covering this case too.

test/compiler.test.js
```diff
@@ -28,8 +28,34 @@ describe('compiler', function() {
...
+            },
+            sourceFile: path.join(SIMPLE_APP, 'common', 'models', 'foo-bar-bat-baz-with-def!!!.js')
+          },
+          {
+            definition: {
+              name: 'i-should-not-exist' //not in 'models'
+            },
+            sourceFile: path.join(SIMPLE_APP, 'common', 'models', 'i-should-not-exist.js')
+          }
+        ],
```
bajtos: This "from options" block is already pretty bad, please don't make it worse.

Write a new standalone test for each different scenario, see the following tests for inspiration: https://github.com/shlomiassaf/loopback-boot/blob/d9d0e0297bf8007649778e25729d4b7d8077921c/test/compiler.test.js#L580-L723

shlomiassaf: Thought about it, but stayed within the original test :)

I will change it.

test/fixtures/simple-app/common/models/foo-bar-bat-baz-with-def.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function (fooBarBatBazWithDef) {
+  fooBarBatBazWithDef.loaded = true;
+};
```
bajtos: As I mentioned above, please rewrite the unit-tests using the newer approach so that application files are created by the test and they are not in fixtures, you can use [this test](https://github.com/shlomiassaf/loopback-boot/blob/d9d0e0297bf8007649778e25729d4b7d8077921c/test/compiler.test.js#L555-L578) as a good example.

bajtos:
@shlomiassaf thank you for the reworked pull request, it looks much better now. Please address my comments above.

I also need you to sign our CLA before I can land the patch. See the details of the failed "clahub" check at the bottom.

shlomiassaf:
@bajtos ,

Addressed your notes.

I pushed them before signing the CLA so the link is lost.
You can supply one or I can wait for the CLAHUB fail.

Thanks.

bajtos:
The link to CLA: https://cla.strongloop.com/agreements/strongloop/loopback-boot

bajtos:
@slnode test please

shlomiassaf:
@bajtos,

Can we land this?

lib/compiler.js
```diff
@@ -263,6 +265,36 @@ function sortByInheritance(instructions) {
...
+function verifyModelDefinitions(rootDir, modelDefinitions) {
+  if (!modelDefinitions || modelDefinitions.length < 1) {
+    return undefined;
+  }
+
+  var registry = {};
+  modelDefinitions.forEach(function(definition, idx) {
+    try {
+      var fullPath = path.resolve(rootDir, definition.sourceFile);
+      definition.sourceFile = require.resolve(fullPath);
```
bajtos: This does not support the case when the `sourceFile` value does not have any extension and there are two files matching this value - one is `.json` and the other `.coffee'. See #59 for more details.

bajtos:
Thank you for the update. AFAICT, you haven't addressed one of my initial comments yet. Plus I found another issue, see the new comment above.

bajtos:
@shlomiassaf ping, what's the status of this patch? Are you still keen to finish it?

shlomiassaf:
@bajtos , sorry for the delay, work pressure.

Will do my best to finish this week, thanks for the update.

lib/compiler.js
shlomiassaf: @bajtos, added.
unit test is not needed, it is tested in "should loads model without definition".

Since the block is in try/catch it will not load it anyway, just waste some resources try to get it... This is the implementation in "loadModelDefinition()"

I blocked it with an if statement just in case...

shlomiassaf:
@bajtos, you can run tests

bajtos:
@slnode ok to test

bajtos:
@shlomiassaf Thank you for the update. I have quite a lot of other work on my plate, so I can't review right now, but I'll try to get back to this as soon as possible.

bajtos:
@shlomiassaf would you mind rebasing your patch on top of the current master branch and fixing the merge conflicts?

bajtos:
@shlomiassaf I have rebase the code on the current master, fixed coding style issues and landed the patch via 2815a42. Thank you for the contribution!

PR #102, opened 2015-03-04
Fix test for different line endings on Windows

PradnyaBaviskar:
The test "configures middleware using shortform" failed on windows due to different line endings. Fixed it.

@bajtos, can you please review it?

### DIFF
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -7,6 +7,7 @@ var fs = require('fs-extra');
 var sandbox = require('./helpers/sandbox');
 var appdir = require('./helpers/appdir');
 var supertest = require('supertest');
+var os = require('os');
 
 var SIMPLE_APP = path.join(__dirname, 'fixtures', 'simple-app');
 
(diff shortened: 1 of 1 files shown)


test/executor.test.js
```diff
@@ -434,9 +435,11 @@ describe('executor', function() {
...
       .end(function(err, res) {
         if (err) return done(err);
-        expect(res.text).to.eql('<!DOCTYPE html>\n<html>\n<head lang="en">\n' +
-          '    <meta charset="UTF-8">\n    <title>simple-app</title>\n' +
-          '</head>\n<body>\n<h1>simple-app</h1>\n</body>\n</html>');
+        expect(res.text).to.eql('<!DOCTYPE html>' + os.EOL + '<html>' + os.EOL +
+          '<head lang="en">' + os.EOL + '' + '    <meta charset="UTF-8">' +
+          os.EOL + '    <title>simple-app</title>' + os.EOL + '</head>' +
+          os.EOL + '<body>' + os.EOL + '<h1>simple-app</h1>' + os.EOL +
+          '</body>' + os.EOL + '</html>');
```
bajtos: I'm wondering if there is any cleaner way how to fix this, so that the text can keep using `\n` characters. How about something like this?

```js
expect(res.text).to.eql('....'.replace(/\n/g, os.EOL));
```

PradnyaBaviskar:
>  How about something like this? 
> ```expect(res.text).to.eql('....'.replace(/\n/g, os.EOL));```

@bajtos , your approach looks more cleaner. Modified the test case to match it. 

Can you please review?

test/executor.test.js
```diff
@@ -434,9 +435,10 @@ describe('executor', function() {
       .get('/')
       .end(function(err, res) {
         if (err) return done(err);
-        expect(res.text).to.eql('<!DOCTYPE html>\n<html>\n<head lang="en">\n' +
+        expect(res.text).to.eql(('<!DOCTYPE html>\n<html>\n<head lang="en">\n' +
```
bajtos: Nitpick: the extra '()' should not be needed.

bajtos:
Landed, thank you.

PR #103, opened 2015-03-05
Add feature to disable component

PradnyaBaviskar:
In reference to https://github.com/strongloop/loopback-explorer/issues/51, added implementation to disable components through ```component-config.json```

@bajtos, please review.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -480,18 +480,20 @@ function resolveMiddlewareParams(rootDir, params) {
 }
 
 function buildComponentInstructions(rootDir, componentConfig) {
-  return Object.keys(componentConfig).map(function(name) {
-    var sourceFile;
-    if (name.indexOf('./') === 0 || name.indexOf('../') === 0) {
-      // Relative path
-      sourceFile = path.resolve(rootDir, name);
-    } else {
-      sourceFile = require.resolve(name);
-    }
+  return Object.keys(componentConfig)
+    .filter(function(name) { return !!componentConfig[name]; })
+    .map(function(name) {
+      var sourceFile;
+      if (name.indexOf('./') === 0 || name.indexOf('../') === 0) {
+        // Relative path
+        sourceFile = path.resolve(rootDir, name);
+      } else {
+        sourceFile = require.resolve(name);
+      }
 
-    return {
-      sourceFile: sourceFile,
-      config: componentConfig[name]
-    };
-  });
+      return {
+        sourceFile: sourceFile,
+        config: componentConfig[name]
+      };
+    });
 }
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -468,7 +468,25 @@ describe('executor', function() {
 
     boot(app, appdir.PATH);
 
+    expect(Object.keys(require.cache)).to.include(
+      appdir.resolve('components/test-component/index.js'));
+
     expect(app.componentOptions).to.eql({ option: 'value' });
+  });
+
+  it('disables component when configuration is not set', function() {
+    appdir.writeConfigFileSync('component-config.json', {
+      './components/test-component': false
+    });
+
+    appdir.writeFileSync('components/test-component/index.js',
+      'module.exports = ' +
+      'function(app, options) { app.componentOptions = options; }');
+
+    boot(app, appdir.PATH);
+
+    expect(Object.keys(require.cache)).to.not.include(
+      appdir.resolve('components/test-component/index.js'));
   });
 
   it('configures middleware (that requires `this`)', function(done) {


lib/executor.js
```diff
@@ -290,6 +290,9 @@ function setupMiddleware(app, instructions) {
 
 function setupComponents(app, instructions) {
   instructions.components.forEach(function(data) {
+    if (data === undefined)
+      return;
+
```
bajtos: Please filter out `undefined` records inside the compiler.

A possible implementation, sorry for giving you a wrong one initially:

```
return Object.keys(...)
  .filter(..check config is truthy..)
  .map(..original map function..)
```

bajtos:
One comment, the rest looks good.

PradnyaBaviskar:
Made following changes
1) Filtered out ```undefined``` components from compiler itself.
2) Modified the test case to verify that disabled component is not loaded

@bajtos , can you please review?

lib/compiler.js
```diff
@@ -480,18 +480,20 @@ function resolveMiddlewareParams(rootDir, params) {
...
-    return {
-      sourceFile: sourceFile,
-      config: componentConfig[name]
-    };
-  });
+      return {
+        sourceFile: sourceFile,
+        config: componentConfig[name]
+      };
+    });
```
bajtos: Are you using tab characters here? Please ensure you use only spaces for indentation (you may need to change the settings in your editor).

test/executor.test.js
```diff
@@ -468,9 +468,25 @@ describe('executor', function() {
 
     boot(app, appdir.PATH);
 
+    expect(containsSubstring(Object.keys(require.cache),
+      'components\\test-component\\index.js')).to.eql(true);
```
bajtos: 1) This will not work on unix, where the path separator is `\`.
2) It's possible that multiple tests are using the same relative path, in which case this check will incorrectly pass. A better solution is to use build a full path using `sandbox.resolve('components', 'test-componen', 'index.js')` and the assert that the keys contain this entry (there is a chai check for that).

test/executor.test.js
```diff
@@ -468,9 +468,25 @@ describe('executor', function() {
...
+      './components/test-component1': false
+    });
+
+    appdir.writeFileSync('components/test-component1/index.js',
+      'module.exports = ' +
+      'function(app, options) { app.componentOptions = options; }');
+
+    boot(app, appdir.PATH);
+    expect(containsSubstring(Object.keys(require.cache),
+      'components\\test-component1\\index.js')).to.eql(false);
```
bajtos: Same comment as above.

lib/compiler.js
PradnyaBaviskar: I've used space character here for indentation. I also cross-verified the editor settings. And it is set to use "spaces" only for indentation. Although not sure, why does it highlight here in green!

PradnyaBaviskar:
Modified the test case to use ```appdir.resolve('components', 'test-componen', 'index.js')```, for verifying that disabled components are not loaded.

@bajtos - can you please review?

lib/compiler.js
bajtos: Perhaps it's GitHub's way of showing that the line differs in whitespace only. Don't worry about that then.

test/executor.test.js
```diff
@@ -468,9 +468,27 @@ describe('executor', function() {
 
     boot(app, appdir.PATH);
 
+    expect(Object.keys(require.cache)).to
+    .include(appdir.resolve('components/test-component/index.js'));
```
bajtos: Style: it's best to keep the chain of chai assertions on a single line. Also when a statement continues on a next line, the next lines should be indented by two spaces. 

```js
// either
expect(Object.keys(require.cache))
  .to.include(appdir.resolve('components/test-component/index.js'));

// or 
expect(Object.keys(require.cache)).to.include(
  appdir.resolve('components/test-component/index.js'));
```

The same comment applies to the test below.

test/executor.test.js
```diff
@@ -468,9 +468,27 @@ describe('executor', function() {
...
+
     expect(app.componentOptions).to.eql({ option: 'value' });
   });
 
+  it('disables component when configuration is not set', function() {
+    appdir.writeConfigFileSync('component-config.json', {
+      './components/test-component1': false
```
bajtos: Now that you are using `appdir.resolve`, is it still needed to use a unique name instead of `./components/test-component`?

bajtos:
Also please sign the CLA for this project here: https://cla.strongloop.com/agreements/strongloop/loopback-boot (the page is linked from "Details" in the list of failing checks too).

PradnyaBaviskar:
@bajtos,
Made suggested changes. Can you review?

bajtos:
@slnode test please

bajtos:
Landed, thank you.

Issue #65, opened 2014-11-10
Relative paths in `options.bootScripts`


bajtos:
#31 introduced `options.bootScripts` that can be used to specify additional boot scripts to execute. The script paths are passed directly to `require`, which means that relative paths are resolved relatively to `node_modules/loopback-boot/lib`.

We should improve the compiler and resolve any relative paths using `appRootDir`:

```
/absolute/path.js -> /absolute/path/.js
module/relative.js -> module/relative.js
./app/relative -> {appRootDir}/app/relative.js
```

A similar problem in the way how modules, middleware and components are resolved is tracked by #73.

bajtos:
@PradnyaBaviskar I believe the problem should be reproducible by modifying an existing unit-test in [compiler.test.js](https://github.com/strongloop/loopback-boot/blob/bfcd3ff15e35293b4c4497fc3d01c6ed164d68b5/test/compiler.test.js#L397-L406).

```js
    it('supports `bootScripts` option', function() {
      appdir.createConfigFilesSync();
      var initJs = appdir.writeFileSync('custom-boot/init.js',
        'module.exports = function(app) { app.fnCalled = true; };');
      var instructions = boot.compile({
        appRootDir: appdir.PATH,
        bootScripts: ['./custom-boot/init.js'] // << MODIFIED
      });
      expect(instructions.files.boot).to.eql([initJs]);
    });
```

bajtos:
Fixed by #104. Note that module paths are still resolved against loopback-boot node_modules, see #73.

Issue #73, opened 2014-11-21
Resolve module paths as relative to appRootDir instead of `loopback-boot/lib`


raymondfeng:
In some cases, we need to run an embedded sandbox application during tests so that `require` will be based on the root of the embedded application instead of the host module.

For example,

```
loopback-boot
  |_ test/fixure/test-app
                       |_ package.json
                       |_node_modules
                          |_test-module
```
See:

https://github.com/strongloop/loopback-boot/pull/70
https://github.com/strongloop/loopback/issues/833

bajtos:
Typically, both loopback-boot and the module used from boot are sitting in the same node_modules folder:

```
project/
  node_modules/
    loopback-boot/
    boot-dependency/
```

When `require.resolve` does not find "boot-dependency" in `loopback-boot/node_modules`, it traverses the parent directories and eventually finds the module in `project/node_modules/boot-dependency`.

In order to reproduce the situation described by this GH issue, one can create the following file/directory structure in a unit-tests:

```
looopback-boot/
  sandbox/
   {app-uid}/
    node_modules/
      bootdep/
        index.js
        package.json # possibly not needed
```

And then reference "bootdep" from model sources, components and/or middleware config, e.g.

```
// sandbox/{app-uid}/middleware.json
{
  "initial": {
    "bootdep": { /*empty config */ }
  }
}
```

bajtos:
@PradnyaBaviskar I think this issue should be fixed by your linked pull requests now, is that correct?

PradnyaBaviskar:
Yes @bajtos, this would now be fixed with my linked PL.

PR #104, opened 2015-03-10
Resolve relative paths in bootScripts using appRootDir

PradnyaBaviskar:
In reference to - https://github.com/strongloop/loopback-boot/issues/65
1. Added function to resolve relative paths in both ```bootScripts``` and ```bootDirs``` with respect to the app root directory.
2. Added test cases to verify the change.

@bajtos, can you please review?

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -62,8 +62,11 @@ module.exports = function compile(options) {
   // require directories
   var bootDirs = options.bootDirs || []; // precedence
   bootDirs = bootDirs.concat(path.join(appRootDir, 'boot'));
+  resolveRelativePaths(bootDirs, appRootDir);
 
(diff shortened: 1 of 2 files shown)


bajtos:
@PradnyaBaviskar this is a good start. However, you are not handling the case where `module/relative.js` should be resolved to `appRootDir/node_modules/module/relative.js` (or the place where `require.resolve` would look for that file if it was required from the application). I believe that makes this PR a breaking change.

I am proposing to add a check to your new code and resolve the path only if it starts with `./` or `../`. 

The proper way of resolving module paths can be deferred for #73.

PradnyaBaviskar:
Done - Added check to resolve path only if it starts with ```./``` or ```..```

@bajtos - can you review?

bajtos:
Landed, thank you.

Issue #105, opened 2015-03-12
Cannot boot two browserified apps


ritch:
https://github.com/strongloop/loopback-boot/blob/bc233e83d33ae29ab2ffb75fe9798d0b01388db8/browser.js#L20

Uses a single global name to reference the `instructions`. You should be able to specify this in `boot.compileToBrowserify()` and `boot()`.

ritch:
@bajtos

bajtos:
How do you propose to handle multiple loopback instances, especially multiple instances of the global model registry when there are multiple application loaded?

ritch:
> How do you propose to handle multiple loopback instances, especially multiple instances of the global model registry when there are multiple application loaded?

Isn't this out of scope for this issue? 

See https://github.com/strongloop/loopback/issues/1056 for my take on this.

bajtos:
> Isn't this out of scope for this issue?

It's a prerequisite. See also https://github.com/strongloop/loopback-datasource-juggler/issues/119 - "The model names have to be unique."

PR #106, opened 2015-03-13
Support for multiple apps in browserified bundle.

kraman:
/to @bajtos
/cc @ritch 
Connect to #105

### DIFF
--- a/index.js
+++ b/index.js
@@ -132,6 +132,9 @@ exports = module.exports = function bootLoopBackApp(app, options, callback) {
 /**
  * Compile boot instructions and add them to a browserify bundler.
  * @param {Object|String} options as described in `bootLoopBackApp` above.
+ * @property {String} [appId] Application identifier used to load the correct
+ * boot configuration when building multiple applications using browserify.
+ * @end
  * @param {Object} bundler A browserify bundler created by `browserify()`.
  *
  * @header boot.compileToBrowserify(options, bundler)
--- a/lib/bundler.js
+++ b/lib/bundler.js
@@ -91,11 +91,19 @@ function bundleInstructions(instructions, bundler) {
    b.require(instructionsStream, { expose: 'loopback-boot#instructions' });
    */
 
+  var instructionId = 'instructions';
+  // Create an unique instruction identifier using the application ID.
+  // This is only useful when multiple loopback applications are being bundled
+  // together.
+  if (instructions.appId)
+    instructionId += '-' + instructions.appId;
+
   // Write the instructions to a file in our node_modules folder.
   // The location should not really matter as long as it is .gitignore-ed
   var instructionsFile = path.resolve(__dirname,
-    '..', 'generated-instructions.json');
-
+    '..', 'generated-' + instructionId + '.json');
   fs.writeFileSync(instructionsFile, instructionsString, 'utf-8');
-  bundler.require(instructionsFile, { expose: 'loopback-boot#instructions' });
+
+  var moduleName = 'loopback-boot#' + instructionId;
+  bundler.require(instructionsFile, { expose: moduleName });
 }
--- a/test/browser.multiapp.test.js
+++ b/test/browser.multiapp.test.js
@@ -1,0 +1,102 @@
+var boot = require('../');
+var exportBrowserifyToFile = require('./helpers/browserify').exportToSandbox;
+var fs = require('fs');
+var path = require('path');
+var expect = require('chai').expect;
+var browserify = require('browserify');
+var sandbox = require('./helpers/sandbox');
+var vm = require('vm');
+var createBrowserLikeContext = require('./helpers/browser').createContext;
+var printContextLogs = require('./helpers/browser').printContextLogs;
+
+describe('browser support for multiple apps', function() {
+  this.timeout(60000); // 60s to give browserify enough time to finish
+
+  beforeEach(sandbox.reset);
+
+  it('has API for bundling and booting multiple apps', function(done) {
+    var app1Dir = path.resolve(__dirname, './fixtures/browser-app');
+    var app2Dir = path.resolve(__dirname, './fixtures/browser-app-2');
+
+    var apps = [
+      {
+        appDir: app1Dir,
+        appFile: './app.js',
+        moduleName: 'browser-app'
+      },
+      {
+        appDir: app2Dir,
+        appFile: './app.js',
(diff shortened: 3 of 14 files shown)


bajtos:
The implementation looks good in general. I would use `options.appId` or something else that is more descriptive than `options.id`. I need you @kraman to add unit-tests before this can be landed.

bajtos:
And also not that the CI build is failing.

kraman:
@slnode test please

kraman:
@bajtos @ritch PTAL

index.js
```diff
@@ -132,6 +132,9 @@ exports = module.exports = function bootLoopBackApp(app, options, callback) {
 /**
  * Compile boot instructions and add them to a browserify bundler.
  * @param {Object|String} options as described in `bootLoopBackApp` above.
+ * @property {String} [appId] Application identifier used to load the correct
```
bajtos: Are we planning to use this id to distinguish between applications at runtime? If not, then I am proposing to use `bundleId` as the option name.

lib/bundler.js
```diff
@@ -91,11 +91,19 @@ function bundleInstructions(instructions, bundler) {
    b.require(instructionsStream, { expose: 'loopback-boot#instructions' });
    */
 
+  var instructionId = 'instructions';
+  // Create an unique instruction identifier using the application ID.
+  // This is only useful when multiple loopback applications are being bundled
+  // together.
+  if (instructions.applicationId)
```
bajtos: The docs says `appId`, this code is using `applicationId`, is that intended?

As mentioned in my comment above, I'd like to use `bundleId` instead.

bajtos: I see, you are converting `appId` to `applicationId` below. Ignore the first part of the comment then.

bajtos:
Please refactor the test implementation so that the existing tests can remain unchanged.

test/fixtures/browser-app-2/model-config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "_meta": {
+    "sources": [
+      "./models",
+      "loopback/common/models"
+    ]
+  },
+  "Email": {
```
bajtos: Email is a built-in loopback model, pick a different name please.

test/browser.test.js
```diff
@@ -77,40 +74,96 @@ describe('browser support', function() {
...
+      var app1 = bundledApps.defaultApp;
+      var app2 = bundledApps.browserApp2;
 
-  boot.compileToBrowserify(appDir, b);
+      expect(app1.settings).to.have.property('custom-key', 'custom-value');
+      expect(Object.keys(app1.models)).to.include('Customer');
+      expect(app1.models.Customer.settings)
+        .to.have.property('_customized', 'Customer');
+
+      expect(Object.keys(app2.models)).to.include('Email');
```
bajtos: Should this test also verify that `app1` does not contain app2's model and `app2` does not contain `Customer` model?

index.js
kraman: @bajtos Depends what you call an app vs a bundle. 

To me this is distinguishing between apps. You can have multiple instance of the app use the same browserified instructions. 

Two distinct apps would probably not use the same browserified instructions.

I would keep this `appId`.

kraman:
Re:

> Please refactor the test implementation so that the existing tests can remain unchanged.

Do you have ideas how I can do that? The compile strategies were creating their own `browserify` instances which cannot be reused. I had to refactor them in order to package multiple apps in the same bundle.

test/fixtures/browser-app-2/model-config.json
kraman: Sure, I'll pick a different name.

test/browser.test.js
kraman: done

kraman:
@bajtos PTAL

index.js
bajtos: Fair enough.

lib/bundler.js
bajtos: Is it worth using `appId` instead of `applicationId` here for the sake of consistency?

test/browser.multiapp.test.js
```diff
@@ -0,0 +1,100 @@
...
+        'Customer');
+
+      expect(Object.keys(app2.models)).to.include('Robot');
+      expect(Object.keys(app2.models)).to.not.include('Customer');
+
+      done();
+    });
+  });
+});
+function browserifyTestApp(apps, next) {
```
bajtos: `browserifyTestApps` (plural), add an empty line between L54-L55 to visually delimit functions.

test/browser.multiapp.test.js
```diff
@@ -0,0 +1,100 @@
...
+    var moduleName = apps[i].moduleName;
+    var appId = apps[i].appId;
+
+    appFile = path.join(appDir, appFile);
+    b.require(appFile, {expose: moduleName});
+
+    if (appId) {
+      boot.compileToBrowserify({
+        appId: appId, appRootDir: appDir
+      }, b);
```
bajtos: Style: place each function arg on a new line.

```js
boot.compileToBrowserify(
  { appId: appId, appRootDir: appDir },
  b);
```

Or build the options upfront:

```js
var opts = appDir;
if (appId) {
  opts = { appId: appId, appRootDir: appDir }
}
boot.compileToBrowserify(opts, b);
```

test/browser.multiapp.test.js
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+      boot.compileToBrowserify(appDir, b);
+    }
+  }
+
+  exportBrowserifyToFile(b, 'browser-app-bundle.js', next);
+}
+
+function executeBundledApp(bundlePath, apps) {
```
bajtos: Plural - `executeBundledApps`.

test/helpers/browserify.js
```diff
@@ -0,0 +1,16 @@
+var fs = require('fs');
+var sandbox = require('./sandbox');
+
+function exportToFile(b, fileName, callback) {
```
bajtos: Since you are resolving `fileName` to a path in sandbox, how about `exportToSandbox` or `bundleToSandbox`?

bajtos:
LGTM. Please wait for the CI build to pass before landing.

kraman:
@slnode test please

PR #107, opened 2015-03-13
Resolve module paths as relative to appRootDir

PradnyaBaviskar:
In context of - https://github.com/strongloop/loopback-boot/issues/73, made following changes...
 1. Added test case to verify that a call to <br/>
```app.boot(app, { appRootDir: __dirname, bootScripts: ["./custom-boot/script" ] })```
<br/> resolves boot script as ``` __dirname+"/custom-boot/script.js"```

@bajtos, please review.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -150,7 +150,7 @@ function findScripts(dir) {
   var results = [];
   files.forEach(function(filename) {
     // ignore index.js and files prefixed with underscore
-    if ((filename === 'index.js') || (filename[0] === '_')) {
+    if (filename === 'index.js' || filename[0] === '_') {
       return;
     }
 
@@ -165,11 +165,7 @@ function findScripts(dir) {
       else
         debug('Skipping file %s - unknown extension', filepath);
     } else {
-      try {
-        path.join(require.resolve(filepath));
-      } catch (err) {
-        debug('Skipping directory %s - %s', filepath, err.code || err);
-      }
+      debug('Skipping directory %s', filepath);
     }
   });
 
@@ -270,7 +266,7 @@ function findModelDefinitions(rootDir, sources) {
   var registry = {};
 
   sources.forEach(function(src) {
-    var srcDir = resolveSourceDir(rootDir, src);
+    var srcDir = resolveAppPath(rootDir, src);
     if (!srcDir) {
       debug('Skipping unknown module source dir %j', src);
       return;
@@ -298,14 +294,15 @@ function findModelDefinitions(rootDir, sources) {
   return registry;
 }
 
-function resolveSourceDir(rootDir, sourceDir) {
-  var srcDir = path.resolve(rootDir, sourceDir);
-  if (fs.existsSync(srcDir))
-    return srcDir;
+function resolveAppPath(rootDir, relativePath) {
+  var fullPath = path.resolve(rootDir, relativePath);
+  if (fs.existsSync(fullPath))
+    return fullPath;
 
-  // Handle module-relative path, e.g. `loopback/common/models`
-  var start = sourceDir.substring(0, 2);
+  var start = relativePath.substring(0, 2);
   if (start !== './' && start !== '..') {
+    // Handle module-relative path, e.g. `loopback/common/models`
+
     // Module.globalPaths is a list of globally configured paths like
     //   [ env.NODE_PATH values, $HOME/.node_modules, etc. ]
     // Module._nodeModulePaths(rootDir) returns a list of paths like
@@ -313,16 +310,24 @@ function resolveSourceDir(rootDir, sourceDir) {
     var modulePaths = Module.globalPaths
       .concat(Module._nodeModulePaths(rootDir));
 
-    srcDir = modulePaths
+    fullPath = modulePaths
       .map(function(candidateDir) {
-        return path.join(candidateDir, sourceDir);
+        return path.join(candidateDir, relativePath);
       })
       .filter(function(candidate) {
         return fs.existsSync(candidate);
       })
       [0];
-    if (srcDir)
-      return srcDir;
+    if (fullPath)
+      return fullPath;
+  } else {
+    // Handle relative path, e.g. `./common/models`
+    try {
+      fullPath = require.resolve(fullPath);
+      return fullPath;
+    } catch (err) {
+      debug ('Skipping %s - %s', fullPath, err);
+    }
   }
 
   return undefined;
@@ -503,9 +508,11 @@ function buildComponentInstructions(rootDir, componentConfig) {
 
 function resolveRelativePaths(relativePaths, appRootDir) {
   relativePaths.forEach(function(relativePath, k) {
-    var start = relativePath.substring(0, 2);
-    if (start === './' || start === '..') {
-      relativePaths[k] = path.resolve(appRootDir, relativePath);
+    var resolvedPath = resolveAppPath(appRootDir, relativePath);
+    if (resolvedPath !== undefined) {
+      relativePaths[k] = resolvedPath;
+    } else {
+      debug ('skipping boot script %s - unknown file', relativePath);
     }
   });
 }
--- a/test/compiler.test.js
+++ b/test/compiler.test.js
@@ -439,6 +439,69 @@ describe('compiler', function() {
       expect(instructions.files.boot).to.eql([initJs]);
     });
 
+    it('should resolve non-relative path in `bootScripts`', function() {
+      appdir.createConfigFilesSync();
+      var initJs = appdir.writeFileSync('custom-boot/init.js', '');
+      var instructions = boot.compile({
+        appRootDir: appdir.PATH,
+        bootScripts: ['custom-boot/init.js']
+      });
+      expect(instructions.files.boot).to.eql([initJs]);
+    });
+
+    it('should resolve non-relative path in `bootDirs`', function() {
+      appdir.createConfigFilesSync();
+      var initJs = appdir.writeFileSync('custom-boot/init.js', '');
+      var instructions = boot.compile({
+        appRootDir: appdir.PATH,
+        bootDirs:['custom-boot']
+      });
+      expect(instructions.files.boot).to.eql([initJs]);
+    });
+
+    it('resolves missing extensions in `bootScripts`', function() {
+      appdir.createConfigFilesSync();
+      var initJs = appdir.writeFileSync('custom-boot/init.js', '');
+      var instructions = boot.compile({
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lib/compiler.js [hunk 1]
```diff
@@ -323,8 +319,15 @@ function resolveSourceDir(rootDir, sourceDir) {
       [0];
     if (srcDir)
       return srcDir;
+  } else {
+    try {
+      var resolved = path.resolve(relativePath);
```
bajtos: Please give `resolved` a better, more descriptive name. E.g. `absolutePath` or `fullPath`.

Add a comment explaining why we need to call `require.resolve`.

lib/compiler.js
```diff
@@ -298,13 +294,13 @@ function findModelDefinitions(rootDir, sources) {
...
 }
 
-function resolveSourceDir(rootDir, sourceDir) {
-  var srcDir = path.resolve(rootDir, sourceDir);
+function resolveAppPath(rootDir, relativePath) {
+  var srcDir = path.resolve(rootDir, relativePath);
   if (fs.existsSync(srcDir))
     return srcDir;
 
   // Handle module-relative path, e.g. `loopback/common/models`
```
bajtos: This comment feels misleading now, because the code below handles both relative paths and module-relative paths now. I am proposing to move the comment to the "if" block below, and add a comment to the "else" block too.

test/compiler.test.js
```diff
@@ -405,6 +405,17 @@ describe('compiler', function() {
       expect(instructions.files.boot).to.eql([initJs]);
     });
 
+    it('should resolve js files in `bootScripts`', function() {
```
bajtos: The test name IMO does not describe the scenario well. The only relevant distinction of this test is that the bootScript file does not provide extension, right? My proposal: "resolves missing file extensions in `bootScritps`", but feel free to come up with your own name.

In the modules I start from scratch, I prefer to use the present tense in test names instead of starting each test name with `should` - [DRY](http://en.wikipedia.org/wiki/Don%27t_repeat_yourself). (Even though older loopback modules use the "should" convention.) Please keep the test names consistent with the convention used in each project.

lib/compiler.js
```diff
@@ -313,18 +310,25 @@ function resolveSourceDir(rootDir, sourceDir) {
...
       .filter(function(candidate) {
         return fs.existsSync(candidate);
       })
       [0];
-    if (srcDir)
-      return srcDir;
+    if (fullPath)
+      return fullPath;
+  } else {
+    // Handle relative path, e.g. `{appRootDir}/common/models`
```
bajtos: Nitpick: the relative path is usually specified as "./common/models"

lib/compiler.js
```diff
@@ -313,18 +310,25 @@ function resolveSourceDir(rootDir, sourceDir) {
...
+    try {
+      fullPath = require.resolve(fullPath);
+      return fullPath;
+    } catch (err) {
+      debug ('Skipping %s - %s', fullPath, err);
+    }
   }
-
```
bajtos: Nitpick: please keep the empty line as a visual delimiter of different code blocks.

lib/compiler.js
```diff
@@ -505,7 +509,12 @@ function resolveRelativePaths(relativePaths, appRootDir) {
   relativePaths.forEach(function(relativePath, k) {
     var start = relativePath.substring(0, 2);
     if (start === './' || start === '..') {
-      relativePaths[k] = path.resolve(appRootDir, relativePath);
+      var resolvedPath = resolveAppPath(appRootDir, relativePath);
+      if (resolvedPath !== undefined) {
+        relativePaths[k] = resolvedPath;
+      } else {
+        debug ('skipping boot script %s - unknown file', relativePath);
+      }
```
bajtos: Is this covered by a unit test too?

Also I think you can remove the check `if (start === './' || start === '..') {`now, because `resolveAppPath` correctly handles non-relative paths too, or not? Please add a unit-test for this too.

lib/compiler.js
```diff
@@ -149,8 +149,8 @@ function findScripts(dir) {
 
   var results = [];
   files.forEach(function(filename) {
-    // ignore index.js and files prefixed with underscore
-    if ((filename === 'index.js') || (filename[0] === '_')) {
+    // ignore files prefixed with underscore
```
PradnyaBaviskar: When looking for boot scripts in boot directories, the previous code, had ignored ```index.js```. Now that we resolve relative paths of boot directories, before searching for boot scripts, I wonder, if we still must be ignoring ```index.js```. 

If, per the previous code, ```index.js``` is ignored here, then for a scenario like 
```app.boot(app, { appRootDir: __dirname, bootScripts: ["./custom-boot/script" ] })```, where ```script``` is a directory, ```findScripts()``` won't be able to explore ```__dirname+"/custom-boot/script/index.js"```

So, I have modified code here to stop  ignoring ```index.js```.

@bajtos - let me know otherwise.

Comment (bajtos) [inline lib/compiler.js]: I guess it's ok to include `custom-boot/script/index.js`. However, we should keep ignoring first-level index.js files like `custom-boot/index.js`, `boot/index.js`, etc. in order to preserve backward compatibility.

**Comment by bajtos** [comment]
Please add tests for the following two scenarios a fix the implementation if needed.

```js
// 1)
appdir.writeFileSync('node_modules/custom-boot/init.js', '...');
instructions.bootScripts = 'custom-boot/init.js';
// -> file is resolved to node_modules/custom-boot/init.js

// 2)
appdir.writeFileSync('./custom-boot/init.js', '...');
appdir.writeFileSync('node_modules/custom-boot/init.js', '...');
instructions.bootScripts = 'custom-boot/init.js';
// -> file is resolved to {appRootDir}/custom-boot/init.js
// This is not great, but needed to preserve backwards compatibility
```

lib/compiler.js
bajtos: I think you may have a flaw in your reasoning. In the example you described, `findScripts` is called on `boot` only, the entry `./custom-boot/scripts` from `bootScripts` is passed directly to `resolveRelativePaths` as far as I can tell.

PradnyaBaviskar:
Done - Added 2 test cases.

@bajtos  - Can you review?

test/compiler.test.js
```diff
@@ -439,6 +439,86 @@ describe('compiler', function() {
...
+    it('explores index.js in `bootDirs`', function() {
+      appdir.createConfigFilesSync();
+      var initJs = appdir.writeFileSync('custom-boot/index.js',
+        'module.exports = function(app) { app.fnCalled = true; };');
+      var instructions = boot.compile({
+        appRootDir: appdir.PATH,
+        bootDirs:['./custom-boot']
+      });
+      expect(instructions.files.boot).to.eql([initJs]);
+    });
```
bajtos: The index files in the directories listed in `bootDirs` must be skipped, that's the current behaviour. Please rework the test and the implementation to preserve backward compatibility.

test/compiler.test.js
```diff
@@ -439,6 +439,86 @@ describe('compiler', function() {
...
+    it('resolves module relative path in `bootScripts`', function() {
+      appdir.createConfigFilesSync();
+      var initJs = appdir.writeFileSync('node_modules/custom-boot/init.js',
+        'module.exports = function(app) { app.fnCalled = true; };');
+      var instructions = boot.compile({
+        appRootDir: appdir.PATH,
+        bootScripts: ['custom-boot/init.js']
+      });
+      expect(instructions.files.boot).to.eql([initJs]);
+    });
```
bajtos: This tests seems to be the same as L486-L495, am I missing somthing?

test/compiler.test.js
```diff
@@ -439,6 +439,86 @@ describe('compiler', function() {
...
+      var appJs = appdir.writeFileSync('./custom-boot/init.js',
+        'module.exports = function(app) { app.fnCalled = true; };');
+      var moduleJs = appdir.writeFileSync('node_modules/custom-boot/init.js',
+        'module.exports = function(app) { app.fnCalled = true; };');
+      var instructions = boot.compile({
+        appRootDir: appdir.PATH,
+        bootScripts: ['custom-boot/init.js']
+      });
+      expect(instructions.files.boot).to.not.eql([moduleJs]);
+      expect(instructions.files.boot).to.eql([appJs]);
```
bajtos: Please keep only the second assertion (expectation) line. When the array has only a single item "appJs", then it cannot be an array with a single item "moduleJs", so it's not needed to test that.

PradnyaBaviskar:
@bajtos,
Made corrections, per above comments!

Can you review now?

lib/compiler.js
```diff
@@ -149,8 +149,8 @@ function findScripts(dir) {
...
+    if (filename === 'index.js' || filename[0] === '_') {
```
bajtos: Please fix the comment on L152 too.

test/compiler.test.js
```diff
@@ -439,6 +439,76 @@ describe('compiler', function() {
...
+      var initJs = appdir.writeFileSync('custom-boot/init.js',
+        'module.exports = function(app) { app.fnCalled = true; };');
+      var instructions = boot.compile({
+        appRootDir: appdir.PATH,
+        bootScripts:['./custom-boot/init']
+      });
+      expect(instructions.files.boot).to.eql([initJs]);
+    });
+
+    it('fails to explore index.js in `bootDirs`', function() {
```
bajtos: Perhaps "it('ignores index.js in `bootDirs`')" would be a better name?

bajtos:
Two more minor comments, the patch is otherwise good to be merged.

I found one more edge case that is not covered by the tests now:

```js
     var initJs = appdir.writeFileSync('node_modules/custom-boot/init.js',
       'module.exports = function(app) { app.fnCalled = true; };');
     var instructions = boot.compile({
       appRootDir: appdir.PATH,
       bootScripts: ['custom-boot/init']  // missing extension!
     });
     expect(instructions.files.boot).to.eql([initJs]);
```

I am ok to leave that out for a new pull request though.

bajtos:
Could you please simplify the new unit-tests you are adding and create the boot scripts with empty content? They are never loaded/executed, so it does not matter what code is inside.

```diff
- appdir.writeFileSync('custom-boot/init.js',
-   'module.exports = function(app) { app.fnCalled = true; };');
+ appdir.writeFileSync('custom-boot/init.js', '');
```

PradnyaBaviskar:
@bajtos,
Modified the test cases, so that boot scripts are loaded with empty contents. 

However I haven't yet included the edge test case, explained by you above.

Can you please review?

bajtos:
Landed, thank you!

PR #110, opened 2015-03-18
Add test case to verify module path is resolved relative to appRootDir - for models

PradnyaBaviskar:
In continuation with https://github.com/strongloop/loopback-boot/issues/73, 
For models - added test case to ensure that module paths for models are also resolved relative to appRootDir. 

@bajtos - Please review.

### DIFF
--- a/test/compiler.test.js
+++ b/test/compiler.test.js
@@ -613,10 +613,60 @@ describe('compiler', function() {
       });
     });
 
+    it('resolves non-relative path in `modelSources` option', function() {
+      appdir.createConfigFilesSync({}, {}, {
+        Car: { dataSource: 'db' }
+      });
+      appdir.writeConfigFileSync('custom-models/car.json', { name: 'Car' });
+      appdir.writeFileSync('custom-models/car.js', '');
+
+      var instructions = boot.compile({
+        appRootDir: appdir.PATH,
+        modelSources: ['custom-models']
+      });
+
+      expect(instructions.models).to.have.length(1);
+      expect(instructions.models[0]).to.eql({
+        name: 'Car',
+        config: {
+          dataSource: 'db'
+        },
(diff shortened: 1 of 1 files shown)


test/compiler.test.js
```diff
@@ -638,6 +663,31 @@ describe('compiler', function() {
...
+        name: 'Car',
+        config: {
+          dataSource: 'db'
+        },
+        definition: {
+          name: 'Car'
+        },
+        sourceFile: path.resolve(appdir.PATH, 'custom-models', 'car.js')
+      });
+    });
```
bajtos: Since we care about the resolved path only, can you please simplify the second check to the following?

```js
expect(instructions.models[0].sourceFile).to.equal(appdir.resolve('custom-models', 'car.js'));
```

test/compiler.test.js
```diff
@@ -613,6 +613,31 @@ describe('compiler', function() {
...
+      expect(instructions.models[0]).to.eql({
+        name: 'Car',
+        config: {
+          dataSource: 'db'
+        },
+        definition: {
+          name: 'Car'
+        },
+        sourceFile: path.resolve(appdir.PATH, 'custom-models', 'car.js')
+      });
```
bajtos: Same comment here, since we care about the resolved path only, can you please simplify the second check to the following?

```js
expect(instructions.models[0].sourceFile).to.equal(appdir.resolve('custom-models', 'car.js'));
```

bajtos:
@PradnyaBaviskar what is the status of this pull request? If my understanding is correct, then this PR can be closed, because the two new tests are already included in #114?

PradnyaBaviskar:
@bajtos, you are right! I've moved these 2 tests to PL #114 

This PR can be closed.

Issue #79, opened 2014-12-09
Mixins


bajtos:
LoopBack provides two kinds of mixins which we need to support in loopback-boot too:
 - A regular LoopBack model - mixing process copies all properties and methods to the target model.
 - A function - mixing process calls the function after the model has been attached.

**Mixin as a model - NOT IMPLEMENTED**

loopback-boot already has mechanism for defining arbitrary models, we should reuse this mechanism for defining mixins too. Benefits: developers can apply their knowledge about how to create models to create mixins too. Tools like `yo loopback:model` and Studio will support mixin definitions OOTB.

```
// common/models/timestamp-mixin.json
{
  "name": "TimeStampMixin",
  "properties": {
    "timestamp": "Date"
  }
}

// common/models/timestamp-mixin.js
module.exports = function(TimeStampMixin) {
  // setup auto-update of timestamp on save
}
```

To support this kind of mixins, we need to extend [addAllBaseModels()](https://github.com/strongloop/loopback-boot/blob/176f96e2428486c35668cceda85943229d96f184/lib/compiler.js#L187-L210) and [sortByInheritance()](https://github.com/strongloop/loopback-boot/blob/176f96e2428486c35668cceda85943229d96f184/lib/compiler.js#L220-L243) to take into account mixed-in models in addition to base models. Once that is done, model-like mixins will work both on the server and in the browserified client.

Note that this is can be implemented independently (and should be).

**Mixin as a function**

This is the part where we need to come up with a new convention.

Ultimately, I'd like to see a solution similar to what we have for models, where the compiler understands relations between model definitions and mixin definitions, and can build optimised instructions to load only those mixins that are actually used in the app.

Another thing to consider is the new component architecture as outlined [here](https://github.com/strongloop/loopback-example-component/wiki).

I am trying to come up with a small solution that can be incrementally extended in next PRs.

How about this:

 - `compiler` takes two new options: `mixinScripts` that contains a list of source files to load, `normalizeNames` that defines the algorithm for converting file names into mixin/model names.
 - mixin names are built from file names by the compiler, using `normalizeNames`
 - `executor` is extended to define mixins per instructions from the compiler

Usage:

```js
boot(app, {
  appRootDir: __dirname,
  mixins: [ './mixins/timestamp.js' ]
});
```

The next step is to implement clever auto-loading of mixins from directories, as described earlier. That way the usage would become

```
// explicit
boot(app, {
  appRootDir: __dirname,
[...]

PR #111, opened 2015-03-18
Mixin loading update - #79

fabien:
This implements #79

### DIFF
--- a/index.js
+++ b/index.js
@@ -111,6 +111,8 @@ var addInstructionsToBrowserify = require('./lib/bundler');
  * of `{appRootDir}/middleware.json`
  * @property {Object} [components] Component configuration to use instead
  * of `{appRootDir}/component-config.json`
+ * @property {Array.<String>} [mixinSources] List of directories where to look
+ * for files containing model mixin definitions.
  * @property {Array.<String>} [bootDirs] List of directories where to look
  * for boot scripts.
  * @property {Array.<String>} [bootScripts] List of script files to execute
--- a/index.js
+++ b/index.js
@@ -117,6 +117,8 @@ var addInstructionsToBrowserify = require('./lib/bundler');
  * for boot scripts.
  * @property {Array.<String>} [bootScripts] List of script files to execute
  * on boot.
+ * @property {String|Function|Boolean} [normalization] Mixin normalization
+ * format: false, 'none', 'classify', 'dasherize' - defaults to 'classify'.
  * @end
  * @param {Function} [callback] Callback function.
  *
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -84,9 +84,9 @@ module.exports = function compile(options) {
   var modelInstructions = buildAllModelInstructions(
     modelsRootDir, modelsConfig, modelSources);
 
-  var mixinDirs = options.mixinDirs || modelsMeta.mixins || ['./mixins'];
+  var mixinSources = options.mixinSources || modelsMeta.mixins || ['./mixins'];
   var mixinInstructions =
-    buildAllMixinInstructions(appRootDir, mixinDirs, options);
+    buildAllMixinInstructions(appRootDir, mixinSources, options);
 
   // When executor passes the instruction to loopback methods,
   // loopback modifies the data. Since we are loading the data using `require`,
@@ -104,9 +104,9 @@ module.exports = function compile(options) {
   });
 };
 
-function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
+function buildAllMixinInstructions(appRootDir, mixinSources, options) {
   var files = options.mixins || [];
-  mixinDirs.forEach(function(dir) {
+  mixinSources.forEach(function(dir) {
     dir = path.resolve(appRootDir, dir);
     files = files.concat(findScripts(dir));
   });
@@ -114,7 +114,7 @@ function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
   var mixins = {};
   files.forEach(function(filepath) {
     var ext = path.extname(filepath);
-    var name = normalizeMixinName(path.basename(filepath, ext));
+    var name = normalizeMixinName(path.basename(filepath, ext), options);
     mixins[name] = filepath;
   });
 
@@ -534,8 +534,18 @@ function resolveRelativePaths(relativePaths, appRootDir) {
   });
 }
 
-function normalizeMixinName(str) { // classify names: foo-bar => FooBar
-  str = String(str).replace(/[\W_]/g, ' ').toLowerCase();
-  str = str.replace(/(?:^|\s|-)\S/g, function(c) { return c.toUpperCase(); });
-  return str.replace(/\s/g, '');
+function normalizeMixinName(str, options) {
+  var normalization = options.normalization;
+  if (normalization === false || normalization === 'none') return str;
+  if (normalization === 'dasherize') {
+    str = String(str).replace(/[\W_]/g, ' ').toLowerCase();
+    str = str.replace(/\s+/g, '-');
+  } else if (typeof normalization === 'function') {
+    str = normalization(str);
+  } else { // classify
+    str = String(str).replace(/[\W_]/g, ' ').toLowerCase();
+    str = str.replace(/(?:^|\s|-)\S/g, function(c) { return c.toUpperCase(); });
+    str = str.replace(/\s+/g, '');
+  }
+  return str;
 }
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -60,6 +60,7 @@ module.exports = function compile(options) {
     buildComponentInstructions(componentRootDir, componentConfig);
 
   // require directories
+
   var bootDirs = options.bootDirs || []; // precedence
   bootDirs = bootDirs.concat(path.join(appRootDir, 'boot'));
   resolveRelativePaths(bootDirs, appRootDir);
@@ -68,6 +69,7 @@ module.exports = function compile(options) {
   resolveRelativePaths(bootScripts, appRootDir);
 
   bootDirs.forEach(function(dir) {
+    dir = path.resolve(dir);
     bootScripts = bootScripts.concat(findScripts(dir));
   });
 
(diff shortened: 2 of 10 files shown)


test/fixtures/simple-app/mixins/time-stamps.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "Timestamping",
+  "version": "0.0.1",
```
bajtos: IMO the version should stay in `package.json`, but since this metadata is not used by boot nor loopback, then it does not really matter.

However, in order to make future maintenance less expensive and also to make the tests easier to understand, the test data should include only things that are absolutely necessary for the test to run. 

Please trim down this file down to the minimal version needed by the tests.

test/fixtures/simple-app/mixins/Foo.js
```diff
@@ -0,0 +1,10 @@
+// When you create a named function, its name will be used
+// as the mixin name - alternatively, set mixin.mixinName.
```
bajtos: Please add a test for the case "alternatively, set mixin.mixinName"

bajtos:
Thank you @fabien for the pull request, please add "/to @bajtos" next time you send a PR, so that I know to take a look. 

I did a very quick and superficial review and left few comments for you to address, I need to review this patch thoroughly when I have more time (hopefully next week).

fabien:
@bajtos OK, thanks - I'll await your review next week - then I'll fix any comments

lib/compiler.js
```diff
@@ -60,6 +60,7 @@ module.exports = function compile(options) {
     buildComponentInstructions(componentRootDir, componentConfig);
 
   // require directories
+
```
bajtos: Unrelated whitespace change, please revert

lib/compiler.js
```diff
@@ -68,6 +69,7 @@ module.exports = function compile(options) {
   resolveRelativePaths(bootScripts, appRootDir);
 
   bootDirs.forEach(function(dir) {
+    dir = path.resolve(dir);
```
bajtos: This looks like an unrelated change to me, why is it needed by mixins? Could you please revert?

lib/compiler.js
```diff
@@ -91,12 +97,41 @@ module.exports = function compile(options) {
...
       boot: bootScripts
     }
   });
 };
 
+function buildAllMixinInstructions(appRootDir, mixinSources, options) {
+  var extensions = _.without(_.keys(require.extensions), '.json');
+  var files = options.mixins || [];
+  mixinSources.forEach(function(dir) {
+    dir = path.resolve(appRootDir, dir);
```
bajtos: This should use the recently added method `resolveAppPath` here in order to correctly support sources pointing to modules. See also the way how module sources are resolved in https://github.com/strongloop/loopback-boot/blob/d7c67c803aadd46a514114f449fffabe3edf110e/lib/compiler.js#L274-L280.

Here is what I'd like to see: a shared function that can be used by both `findModelDefinitions` and `buildAllMixinInstructions` for converting a list of source dirs into a list of script files to consider.

It's important to keep the mixin behaviour consistent with models. Few cases to keep in mind:
 - `mixinSources: ['fabien-helpers/mixins']` should scan the directory `node_modules/fabien-helpers/mixins`
 - `mixinSources: ['fabien-helper/server/mixins', fabien-helper/common/mixins']` should not throw even when none of the directories exist. (We can discuss this if you disagree.)

lib/compiler.js
```diff
@@ -91,12 +97,41 @@ module.exports = function compile(options) {
...
+    var metafile = path.join(dir, name + '.json');
+    name = normalizeMixinName(name, options);
+    var meta = {};
+    meta.name = name;
+    if (fs.existsSync(metafile)) {
+      // May overwrite name, not id or filepath
+      _.extend(meta, require(metafile));
+    }
+    meta.filepath = filepath;
+    mixins[name] = meta;
```
bajtos: The key in `mixins` may be different from the `name` provided in the json file. Is this intentional?

I think it will be better to either use `meta.name` as the key, or else register the mixin under both names (the one based on a file name, and the one provided by the json file).

lib/compiler.js
```diff
@@ -509,3 +545,19 @@ function resolveRelativePaths(relativePaths, appRootDir) {
...
+  if (normalization === false || normalization === 'none') return str;
+  if (normalization === 'dasherize') {
+    str = String(str).replace(/[\W_]/g, ' ').toLowerCase();
+    str = str.replace(/\s+/g, '-');
+  } else if (typeof normalization === 'function') {
+    str = normalization(str);
+  } else { // classify
+    str = String(str).replace(/[\W_]/g, ' ').toLowerCase();
+    str = str.replace(/(?:^|\s|-)\S/g, function(c) { return c.toUpperCase(); });
+    str = str.replace(/\s+/g, '');
```
bajtos: Please detect the case when `normalization` is defined and it's not any of the values handled above. You should either throw an error or at least report a warning. Otherwise it's very difficult to troubleshoot the situation when you e.g. make a typo in the normalization value and I don't want to get github issues reported by users in this situation.

bajtos: Also please consider rewriting this using a switch/case + return from each case block, I think the implementation will be simpler/cleaner that way.

```js
if (typeof normalization === 'function') return normalization(str);
switch (normalization) {
  // ...
  case undefined:
  case 'classify':
    // classify & return
  default:
    //report an error
```

How should a `null` value be interpreted? Is it the same as `false`/`'none'`?

test/compiler.test.js
```diff
@@ -706,6 +707,47 @@ describe('compiler', function() {
       instructions = boot.compile(appdir.PATH);
       expect(instructions.config).to.not.have.property('modified');
     });
+
+    it('has a mixinSources option', function() {
+      var options = { appRootDir: SIMPLE_APP, mixinSources: [MIXIN_SOURCES] };
+      var instructions = boot.compile(options);
+      expect(instructions.mixins).to.not.have.property('TimeStamps');
+      expect(instructions.mixins).to.have.property('Other');
```
bajtos: This test is difficult to understand on its own. Why `TimeStamps` should not exist and `Other` should exist?

Please use the helpers `appdir.createConfigFilesSync()`, `appdir.writeConfigFileSync()` and `appdir.writeFileSync()` to build the test scenario from within the test.

test/compiler.test.js
```diff
@@ -706,6 +707,47 @@ describe('compiler', function() {
...
+    });
+
+    it('supports mixin name normalization - classify (default)', function() {
+      var options = { appRootDir: SIMPLE_APP };
+      var instructions = boot.compile(options);
+      expect(instructions.mixins).to.have.property('Example');
+      expect(instructions.mixins).to.have.property('Foo');
+      expect(instructions.mixins).to.have.property('TimeStamps');
```
bajtos: Same comment as above. To avoid repetition in these four similar tests you can create a helper to create a set of named mixin files that can be used as `givenMixinFiles('example', 'foo', 'time-stamps')`. Or perhaps `givenAppWithMixinFiles` that would call `appdir.createConfigFilesSync` too.

```js
it('supports mixin name normalization - classify (default)', function() {
  givenAppWithMixinFiles('example', 'foo', 'time-stamps');

  var options = { appRootDir: appdir.PATH };
  var instructions = boot.compile(options);
  // asserts
});
```

test/fixtures/simple-app/mixins/example.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function(Model, options) {
+
+  Model.exampleMixin = true;
```
bajtos: Either assert in a unit-test that the variable was set by the mixin, or else remove this line. The same comment applies to the other mixin scripts too.

test/executor.test.js
```diff
@@ -271,6 +271,38 @@ describe('executor', function() {
...
+          description: 'The postfix string',
+          default: 'At'
+        }
+      };
+
+      expect(meta.TimeStamps.name).to.equal('Timestamping');
+      expect(meta.TimeStamps).to.have.property('filepath');
+      expect(meta.TimeStamps.description).to.equal(description);
+      expect(meta.TimeStamps.options).to.eql(options);
+    });
```
bajtos: It would be nicer to create the test app from the test, rather than relying on a static fixture in `fixtures/simple-app`. However, in the case of the executor, I can live with simple-app too.

The test is testing too many things, please split it into multiple smaller and focused tests. A minimum is two tests, one testing the registry & meta keys, the other testing meta.TimeStamps.

I don't see how `meta.Foo` is different from `meta.TimeStamps` and how `meta.Foo` checks can fail when `meta.TimeStamps` checks pass. Could you please explain? If the fact that `meta.TimeStamps` checks are passing implies that `meta.Foo` checks pass too, them please remove `meta.Foo` checks as they are redundant.

Also add a comment explaining where are `description` and `options` coming from, or read them programatically in the test.

```js
var meta = require('./fixtures/simple-app/mixins/time-stamps.json');
expect(meta.TimeStamps.description).to.equal(meta.description);
// etc.
```

index.js
```diff
@@ -111,10 +111,14 @@ var addInstructionsToBrowserify = require('./lib/bundler');
  * of `{appRootDir}/middleware.json`
  * @property {Object} [components] Component configuration to use instead
  * of `{appRootDir}/component-config.json`
+ * @property {Array.<String>} [mixinSources] List of directories where to look
+ * for files containing model mixin definitions.
```
bajtos: If I remember our previous discussions correctly, the name `mixinSources` should be used for a list of directories which are scanned only for the mixins that are actually used by some model. As opposed to `mixinDirs`, that is pointing to directories where all files are loaded.

This distinction is very important in order to keep the size of the browser bundle as small as possible.

I am not asking you to implement the full "lazy" loading of mixins if you don't have bandwidth for that now, but I am asking you to rename the config option to keep it consistent with `bootDirs` and `modelSources`.

bajtos:
Reviewed, the general direction is very good :+1: 

Please improve the implementation details per my comments above.

Also add a test to [test/browser.test.js](https://github.com/strongloop/loopback-boot/blob/d7c67c803aadd46a514114f449fffabe3edf110e/test/browser.test.js) to verify that mixins are correctly bundled for the browser too. It should be enough to add a mixin to [test/fixtures/browser-app](https://github.com/strongloop/loopback-boot/tree/d7c67c803aadd46a514114f449fffabe3edf110e/test/fixtures/browser-app) and check that it was correctly loaded.

fabien:
@bajtos would you mind taking a stab at the remaining improvements? You are obviously far more familiar with `loopback-boot` and its implications in the bigger `loopback` ecosystem. I don't mind handling the minor fixes and tests, but I could use some help regarding `resolveAppPath` and such.

I know that many people are hoping to see this landed, so maybe we can split the job? Not because of the work, but to be most efficient regarding the improvements you like to see (and not end up in more discussions, pushes, and comments yet again).

Thanks.

bajtos:
Closing in favour of #120

PR #112, opened 2015-03-19
Support per-application registry of models

bajtos:
A follow-up for https://github.com/strongloop/loopback/pull/1212 implementing support for local per-app model registries.

/to @ritch 
/cc @kraman

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -168,23 +168,24 @@ function setupModels(app, instructions) {
 }
 
 function defineModels(app, instructions) {
+  var registry = app.registry || app.loopback;
   instructions.models.forEach(function(data) {
     var name = data.name;
     var model;
 
     if (!data.definition) {
-      model = app.loopback.getModel(name);
+      model = registry.getModel(name);
       if (!model) {
         throw new Error('Cannot configure unknown model ' + name);
       }
       debug('Configuring existing model %s', name);
     } else if (isBuiltinLoopBackModel(app, data)) {
-      model = app.loopback[name];
-      assert(model, 'app.loopback[model] should have been defined');
+      model = registry.getModel(name);
+      assert(model, 'Built-in model ' + name + ' should have been defined');
       debug('Configuring built-in LoopBack model %s', name);
     } else {
       debug('Creating new model %s %j', name, data.definition);
-      model = app.loopback.createModel(data.definition);
+      model = registry.createModel(data.definition);
       if (data.sourceFile) {
         debug('Loading customization script %s', data.sourceFile);
         var code = require(data.sourceFile);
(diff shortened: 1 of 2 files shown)


lib/executor.js
```diff
@@ -168,23 +168,25 @@ function setupModels(app, instructions) {
...
 
     if (!data.definition) {
-      model = app.loopback.getModel(name);
+      model = registry.getModel(name);
       if (!model) {
         throw new Error('Cannot configure unknown model ' + name);
       }
       debug('Configuring existing model %s', name);
     } else if (isBuiltinLoopBackModel(app, data)) {
+      // FIXME(bajtos) This is probably no longer true
```
ritch: Please just use `registry.getModel(name)`

test/executor.test.js
```diff
@@ -199,10 +218,13 @@ describe('executor', function() {
   it('does not call autoAttach', function() {
     boot.execute(app, dummyInstructions);
 
+    // FIXME(bajtos) "loopback.DataSource" must stay available in loopback
+    var DataSource = loopback.DataSource || loopback.registry.DataSource;
```
ritch: This is no longer necessary. See https://github.com/strongloop/loopback/commit/3c509199cc9377d9d2c57a036d8442ebaab31257

bajtos:
@ritch I'll wait with any changes until  strongloop/loopback#1212 is landed.

bajtos:
@ritch I have updated the PR to use the new flag `localRegistry` and addressed your comments. Please review.

lib/executor.js
```diff
@@ -168,23 +168,24 @@ function setupModels(app, instructions) {
 }
 
 function defineModels(app, instructions) {
+  var registry = app.registry || app.loopback;
```
ritch: maybe add a comment that we can drop this (`app.loopback`) in 3.0?

lib/executor.js
```diff
@@ -168,23 +168,24 @@ function setupModels(app, instructions) {
...
     var model;
 
     if (!data.definition) {
-      model = app.loopback.getModel(name);
+      model = registry.getModel(name);
       if (!model) {
         throw new Error('Cannot configure unknown model ' + name);
       }
       debug('Configuring existing model %s', name);
     } else if (isBuiltinLoopBackModel(app, data)) {
```
ritch: Maybe we should create a new registry for built-in models that stands on its own (eg. `loopback.internal = new Registry()`

ritch:
LGTM :+1:

lib/executor.js
bajtos: So far, loopback-boot was not tied to a specific version of loopback, one could use loopback-boot 2.x with loopback 1.x and vice versa. This feature allows developer to split upgrade of loopback runtime and project structure into two smaller independent steps.

I would like to preserve this feature for as long as the maintenance costs are low. Therefore I don't want to drop this part at the time of LoopBack 3.0 release.

lib/executor.js
bajtos: IMO the internal registry will not work because you can't mix models from multiple registries.

PR #113, opened 2015-03-19
Resolve module paths as relative to appRootDir - for middleware

PradnyaBaviskar:
In continuation with #73, 
 1. Made modifications to middleware, so as to resolve the module paths relative to appRootDir.
 2. Added test cases to verify middleware changes.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -266,7 +266,7 @@ function findModelDefinitions(rootDir, sources) {
   var registry = {};
 
   sources.forEach(function(src) {
-    var srcDir = resolveAppPath(rootDir, src);
+    var srcDir = tryResolveAppPath(rootDir, src);
     if (!srcDir) {
       debug('Skipping unknown module source dir %j', src);
       return;
@@ -295,6 +295,16 @@ function findModelDefinitions(rootDir, sources) {
 }
 
 function resolveAppPath(rootDir, relativePath) {
+  var resolvedPath = tryResolveAppPath(rootDir, relativePath);
+  if (resolvedPath === undefined) {
+    var err = new Error('Cannot resolve path "' + relativePath + '"');
+    err.code = 'PATH_NOT_FOUND';
+    throw err;
+  }
+  return resolvedPath;
+}
+
+function tryResolveAppPath(rootDir, relativePath) {
   var fullPath = path.resolve(rootDir, relativePath);
   if (fs.existsSync(fullPath))
     return fullPath;
@@ -429,6 +439,7 @@ function resolveMiddlewarePath(rootDir, middleware) {
   var segments = middleware.split('#');
   var pathName = segments[0];
   var fragment = segments[1];
+  var middlewarePath = pathName;
 
   if (fragment) {
     resolved.fragment = fragment;
@@ -440,7 +451,7 @@ function resolveMiddlewarePath(rootDir, middleware) {
   }
 
   if (!fragment) {
-    resolved.sourceFile = require.resolve(pathName);
+    resolved.sourceFile = resolveAppPath(rootDir, middlewarePath);
     return resolved;
   }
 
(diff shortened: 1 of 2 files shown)


lib/compiler.js
```diff
@@ -440,7 +442,11 @@ function resolveMiddlewarePath(rootDir, middleware) {
   }
 
   if (!fragment) {
-    resolved.sourceFile = require.resolve(pathName);
+    resolvedPath = resolveAppPath(rootDir, middlewarePath);
+    if (resolvedPath === undefined) {
+      throw new Error(pathName);
```
bajtos: Please provide a descriptive error message that makes it easier for the user to understand what is the problem. See https://github.com/joyent/node/blob/9b534e2e87489d219f0d0cf5be7f9119704fe424/lib/module.js#L337-L341 for an example:

```js
  var filename = Module._findPath(request, paths);
  if (!filename) {
    var err = new Error("Cannot find module '" + request + "'");
    err.code = 'MODULE_NOT_FOUND';
    throw err;
  }
  return filename;
```

Please make the change consistently in all places. Perhaps you can create a method `resolveAppPathOrThrow` that can be used everywhere and keep the code building the exception in one place.

bajtos: Actually a better naming convention is `tryResolveAppPath` for the version that returns `undefined` on when not found, and then `resolveAppPath` that throws an error when not found.

PradnyaBaviskar:
Added function ```resolveAppPath``` that throws an exception and the function ```tryResolveAppPath``` returns ```undefined```

@bajtos  - please review.

test/compiler.test.js
```diff
@@ -1043,27 +1044,47 @@ describe('compiler', function() {
...
+
+      expect(instructions.middleware.middleware).to.have.length(1);
+      expect(instructions.middleware.middleware[0])
+        .to.eql(
+          {
+            sourceFile: path.resolve(appJS),
+            config: {
+              phase: 'routes'
+            }
+          });
```
bajtos: This test cares about `sourceFile` property only, please simplify the assertion. You can use the code from the test above. Also IIRC `appdir.writeFileSync` already returns a full (resolved) path, thus it shouldn't be needed to call `path.resolve` again.

```js
expect(instructions.middleware.middleware[0])
  .to.have.property('sourceFile', appJs);
```

test/compiler.test.js
```diff
@@ -1043,27 +1044,47 @@ describe('compiler', function() {
...
 
-        expect(instructions.middleware.middleware[0]).have.property(
-          'sourceFile',
-          appdir.resolve(HANDLER_FILE));
+      expect(instructions.middleware.middleware[0]).have.property(
+        'sourceFile',
+        appdir.resolve(HANDLER_FILE));
+    });
+
+    it('explores middleware function in appRootDir', function() {
```
bajtos: Please use a more descriptive name, e.g. "prefers appRootDir over node_modules for middleware".

bajtos: Actually, I think this is not correct. If I am reading the code correctly, a middleware path like "my-middleware" was never resolved as a relative path in appRooDir before this change.

test/compiler.test.js
```diff
@@ -1043,27 +1044,47 @@ describe('compiler', function() {
...
+      var appJS = appdir.writeFileSync('./my-middleware.js', '');
+      appdir.writeFileSync('node_modules/my-middleware.js', '');
+      appdir.writeConfigFileSync('middleware.json', {
+        'routes': {
+          './my-middleware': {}
```
bajtos: You should use `'my-middleware' :{}` here, otherwise this test never triggers the path that looks in node_modules.

test/compiler.test.js
bajtos: So the correct test should be either "prefers node_modules over appRootDir" or a write different test called "does not treat module-relative paths as appRootDir-relative".

PradnyaBaviskar:
Added one more test case  ```does not treat module-relative paths as appRootDir-relative```, to ensure that ```'my-middleware' :{}'``` looks into node modules and NOT apprRootDir

@bajtos - please review.

bajtos:
Landed, thank you.

PR #114, opened 2015-03-23
Resolve module paths as relative to appRootDir - for component

PradnyaBaviskar:
In continuation with #73,

Made modifications to component, so as to resolve the module paths relative to appRootDir.
Added test case to verify component changes.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -8,6 +8,8 @@ var debug = require('debug')('loopback:boot:compiler');
 var Module = require('module');
 var _ = require('lodash');
 
+var FILE_EXTENSION_JSON = '.json';
+
 /**
  * Gather all bootstrap-related configuration data and compile it into
  * a single object containing instruction for `boot.execute`.
@@ -160,12 +162,11 @@ function findScripts(dir) {
     }
 
     var filepath = path.resolve(path.join(dir, filename));
-    var ext = path.extname(filename);
     var stats = fs.statSync(filepath);
 
     // only require files supported by require.extensions (.txt .md etc.)
     if (stats.isFile()) {
-      if (ext in require.extensions)
+      if (isPreferredExtension(filename))
         results.push(filepath);
       else
         debug('Skipping file %s - unknown extension', filepath);
@@ -299,8 +300,8 @@ function findModelDefinitions(rootDir, sources) {
   return registry;
 }
 
-function resolveAppPath(rootDir, relativePath) {
-  var resolvedPath = tryResolveAppPath(rootDir, relativePath);
+function resolveAppPath(rootDir, relativePath, resolveOptions) {
+  var resolvedPath = tryResolveAppPath(rootDir, relativePath, resolveOptions);
   if (resolvedPath === undefined) {
     var err = new Error('Cannot resolve path "' + relativePath + '"');
     err.code = 'PATH_NOT_FOUND';
@@ -309,12 +310,27 @@ function resolveAppPath(rootDir, relativePath) {
   return resolvedPath;
 }
 
-function tryResolveAppPath(rootDir, relativePath) {
-  var fullPath = path.resolve(rootDir, relativePath);
-  if (fs.existsSync(fullPath))
+function tryResolveAppPath(rootDir, relativePath, resolveOptions) {
+  var fullPath;
+  var start = relativePath.substring(0, 2);
+
+  /* In order to retain backward compatibility, while resolving
+   * component path, `resolveOptions` parameter is added where
+   * `resolveOptions.strict` = true,
+   * means retain backward compatibility when resolving the path and
+   * `resolveOptions.strict` = false,
+   * does not enforce any such restriction when resolving the path */
+  resolveOptions = resolveOptions || { strict: false };
+
+  if (relativePath[0] === '/') {
+    fullPath = relativePath;
+  } else if (start === './' || start === '..' || !resolveOptions.strict) {
+    fullPath = path.resolve(rootDir, relativePath);
+  }
+
+  if (fullPath && fs.existsSync(fullPath))
     return fullPath;
 
-  var start = relativePath.substring(0, 2);
   if (start !== './' && start !== '..') {
     // Handle module-relative path, e.g. `loopback/common/models`
 
@@ -339,6 +355,7 @@ function tryResolveAppPath(rootDir, relativePath) {
         return fs.existsSync(candidate);
       })
       [0];
+
     if (fullPath)
       return fullPath;
   } else {
@@ -356,27 +373,12 @@ function tryResolveAppPath(rootDir, relativePath) {
 
 function loadModelDefinition(rootDir, jsonFile, allFiles) {
   var definition = require(jsonFile);
-  var basename = path.basename(jsonFile, path.extname(jsonFile));
 
   // find a matching file with a supported extension like `.js` or `.coffee`
-  var base;
-  var ext;
-  var validFileType;
-  var sourceFile = allFiles
-    .filter(function(f) {
-      ext = path.extname(f);
-      base = path.basename(f, ext);
-      validFileType = (ext !== '.node') && (ext !== '.json') &&
-        ((typeof require.extensions[ext]) === 'function');
-      return validFileType && (base === basename);
-    })[0];
+  var sourceFile = fixFileExtension(jsonFile, allFiles, true);
 
-  try {
-    sourceFile = path.join(path.dirname(jsonFile), sourceFile);
-    sourceFile = require.resolve(sourceFile);
-  } catch (err) {
-    debug('Model source code not found: %s - %s', sourceFile, err.code || err);
-    sourceFile = undefined;
+  if (sourceFile === undefined) {
+    debug('Model source code not found: %s', sourceFile);
   }
 
   debug('Found model "%s" - %s %s', definition.name,
@@ -456,7 +458,7 @@ function resolveMiddlewarePath(rootDir, middleware) {
   }
 
   if (!fragment) {
-    resolved.sourceFile = resolveAppPath(rootDir, middlewarePath);
+    resolved.sourceFile = resolveAppScriptPath(rootDir, middlewarePath);
     return resolved;
   }
 
@@ -466,7 +468,7 @@ function resolveMiddlewarePath(rootDir, middleware) {
   // function
   var m = require(pathName);
   if (typeof m[fragment] === 'function') {
-    resolved.sourceFile = resolveAppPath(rootDir, middlewarePath);
+    resolved.sourceFile = resolveAppScriptPath(rootDir, middlewarePath);
     return resolved;
   }
 
@@ -484,7 +486,7 @@ function resolveMiddlewarePath(rootDir, middleware) {
 
   for (var ix in candidates) {
     try {
-      resolved.sourceFile = resolveAppPath(rootDir, candidates[ix]);
+      resolved.sourceFile = resolveAppScriptPath(rootDir, candidates[ix]);
       delete resolved.fragment;
       return resolved;
     }
@@ -513,16 +515,8 @@ function buildComponentInstructions(rootDir, componentConfig) {
   return Object.keys(componentConfig)
     .filter(function(name) { return !!componentConfig[name]; })
     .map(function(name) {
-      var sourceFile;
-      if (name.indexOf('./') === 0 || name.indexOf('../') === 0) {
-        // Relative path
-        sourceFile = path.resolve(rootDir, name);
-      } else {
-        sourceFile = require.resolve(name);
-      }
-
       return {
-        sourceFile: sourceFile,
+        sourceFile: resolveAppScriptPath(rootDir, name, {'strict': true}),
         config: componentConfig[name]
       };
     });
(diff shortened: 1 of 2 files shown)


PradnyaBaviskar:
Also added test cases to verify resolving of relative paths for models. 
(Since the pull request - https://github.com/strongloop/loopback-boot/pull/110, was one commit behind the changes made for middleware, I've merged model related test cases from PL-https://github.com/strongloop/loopback-boot/pull/110 into this PL)

@bajtos - can you review?

PradnyaBaviskar:
Extended this PL to prioritize coffeescripts over json for all parts, viz., 
 1. bootscripts
 2. middleware and
 3. component. 

@bajtos  - can you review?

lib/compiler.js
```diff
@@ -8,6 +8,7 @@ var debug = require('debug')('loopback:boot:compiler');
 var Module = require('module');
 var _ = require('lodash');
 
+var FILE_EXTENSION_JSON = '.json';
```
bajtos: Nitpick: please add an empty line to visually delimit the variable (constant) declaration from the next function.

lib/compiler.js
```diff
@@ -538,3 +548,33 @@ function resolveRelativePaths(relativePaths, appRootDir) {
     }
   });
 }
+
+function isPreferredExtension (filename, includeExtensions, excludeExtensions) {
```
bajtos: [KISS](http://en.wikipedia.org/wiki/KISS_principle) and remove `includeExtensions` and `excludeExtensions` that are not used AFAICT.

lib/compiler.js
```diff
@@ -538,3 +548,33 @@ function resolveRelativePaths(relativePaths, appRootDir) {
...
+  includeExtensions = includeExtensions || Object.keys(require.extensions);
+  excludeExtensions = excludeExtensions || [FILE_EXTENSION_JSON];
+  var ext = path.extname(filename);
+  return ((includeExtensions.indexOf(ext) > -1) &&
+    (excludeExtensions.indexOf(ext) === -1));
```
bajtos: You are converting a fast hashmap access O(1) to a slow array scan O(n). Also considering that the two arguments are not used, I find it suboptimal to convert an already-built lookup table in `require.extensions` to an array.

```js
return (ext in includeExtensions) && !(ext in excludeExtensions);
```

lib/compiler.js
```diff
@@ -402,6 +402,14 @@ function buildMiddlewareInstructions(rootDir, config) {
       allConfigs.forEach(function(config) {
         var resolved = resolveMiddlewarePath(rootDir, middleware);
 
+        // prioritize coffee scripts over json
+        if (!isPreferredExtension(resolved.sourceFile)) {
+          var otherFile = tryOtherExtensions(resolved.sourceFile);
+          if (otherFile !== undefined) {
+            resolved.sourceFile = otherFile;
+          }
```
bajtos: This block is repeated in two places, please extract a new method.

```js
resolved.sourceFile = fixFileExtension(resolved.sourceFile);

function fixFileFileExtension(filepath) {
  if (isPreferredExtension(filepath)) return filepath;
  var otherFile = tryOtherExtensions(filepath);
  return otherFile !== undefined ? otherFile: filepath;
}
```

It may be possible to get rid of `tryOtherExtension` completely and inline it inside the new method, I'll leave that for you to decide.

bajtos: What would be even better: a function called e.g. `resolveAppScriptPath` that would call `resolveAppPath` under the hood and then `fixFileExtension` too. That should reduce even more the amount of changes needed.

Since `resolveMiddlewarePath` is calling `resolveAppPath` under the hood, you can call the new method there too and that way the callers of `resolveMiddlewarePath` don't need to know about these implementation details at all.

test/compiler.test.js
```diff
@@ -514,6 +514,37 @@ describe('compiler', function() {
       expect(instructions.files.boot).to.eql([appJs]);
     });
 
+    it('prefers coffeescript over json in `appRootDir/bootDir`', function() {
+      // add coffee-script to require.extensions
+      require('coffee-script/register');
```
bajtos: Please don't register coffee script in every test. Move these two lines to the top of the test file, where other `require` calls are made.

test/compiler.test.js
```diff
@@ -514,6 +514,37 @@ describe('compiler', function() {
...
+
+      appdir.createConfigFilesSync();
+      var coffee = appdir.writeFileSync('node_modules/custom-boot/init.coffee',
+        '');
+      appdir.writeFileSync('node_modules/custom-boot/init.json', '');
+
+      var instructions = boot.compile({
+        appRootDir: appdir.PATH,
+        bootDirs: ['node_modules/custom-boot']
```
bajtos: This is weird. If you want to test the module-based resolution of paths then you should use `bootDirs: ['custom-boot']` only.

I don't know if the option `bootDirs` supports paths pointing to modules. It's possible that such paths are not supported, in which case this test is not necessary.

Also please keep all `bootDirs` close test together, rather than having them intertwined with other tests (e.g. for `bootScripts`).

test/compiler.test.js
```diff
@@ -662,6 +693,72 @@ describe('compiler', function() {
...
+    it('resolves non-relative path in `modelSources` option', function() {
+      appdir.createConfigFilesSync({}, {}, {
+        Car: { dataSource: 'db' }
+      });
+      appdir.writeConfigFileSync('custom-models/car.json', { name: 'Car' });
+      var appJS = appdir.writeFileSync('custom-models/car.js', '');
+
+      var instructions = boot.compile({
+        appRootDir: appdir.PATH,
+        modelSources: ['custom-models']
```
bajtos: IMO, the primary purpose of this variant is to load `node_modules/custom-models` instead of `appRootDir + '/custom-models`.

When the directory does not exist in `node_modules`, then I guess it's ok to try to resolve the path as relative to `appRootDir`, although I am also ok with the solution where non-relative paths are always resolved as a module path in one of `node_modules` folders.

What I would like you to do: Either rework this test to lookup the files in `node_modules/custom-models/car*`, or add a new test that will verify this scenario.

test/compiler.test.js
```diff
@@ -662,6 +693,72 @@ describe('compiler', function() {
...
+
+      expect(instructions.models).to.have.length(1);
+      expect(instructions.models[0].sourceFile).to.equal(appJS);
+    });
+
+    it('resolves non-relative path in `sources` option in `model-config.json`',
+      function() {
+      appdir.createConfigFilesSync({}, {}, {
+        _meta: {
+          sources: ['custom-models']
```
bajtos: Same comment as above:  Either rework this test to lookup the files in `node_modules/custom-models/car*`, or add a new test that will verify this scenario.

test/compiler.test.js
```diff
@@ -1096,6 +1193,38 @@ describe('compiler', function() {
...
+          './my-middleware': {}
+        }
+      });
+
+      var instructions = boot.compile(appdir.PATH);
+
+      expect(instructions.middleware.middleware).to.have.length(1);
+      expect(instructions.middleware.middleware[0]).have.property(
+        'sourceFile', coffee);
+    });
```
bajtos: Please add another unit-test to verify that coffeescript files are loaded from `node_modules` too, i.e. create `node_modules/my-middleware/index.coffee` (+json) instead of `my-middleware.coffee` (+ json).

test/compiler.test.js
```diff
@@ -1195,6 +1324,47 @@ describe('compiler', function() {
...
+      expect(instructions.components[0]).have.property(
+        'sourceFile', appJS
+      );
+    });
+
+    it('loads component relative to node modules', function() {
+      appdir.writeConfigFileSync('component-config.json', {
+        'component': { }
+      });
+      var appJS = appdir.writeConfigFileSync('node_modules/component.js', '');
```
bajtos: Please use `node_modules/component/index.js`, that's a more typical usage.

test/compiler.test.js
```diff
@@ -1195,6 +1324,47 @@ describe('compiler', function() {
...
+
+      var coffee = appdir.writeFileSync('component.coffee', '');
+      appdir.writeFileSync('component.json', '');
+
+      var instructions = boot.compile(appdir.PATH);
+
+      expect(instructions.components).to.have.length(1);
+      expect(instructions.components[0]).have.property(
+        'sourceFile', coffee);
+    });
```
bajtos: Please another test to verify that `node_modules/component/component.coffee` is correctly picked too when the component-config json contains `component` key.

bajtos:
@PradnyaBaviskar thank you for the update, I left a bunch of comments to address. To summarise, I'd like you to:
 - Clean up the current implementation of `tryOtherExtensions`, create a new function `resolveAppScriptPath` that can be used as a replacement for `resolveAppPath` and which will deal with the coffee script extension as needed.
 - Add more tests, especially for the cases where a value `"mymodule"` should be resolved to `{appRootDir}/node_modules/mymodule/index.js` and `{appRootDir}/node_modules/mymodule/index.coffee`.

PradnyaBaviskar:
@bajtos - Addressed all your comments above.

Can you please review?

lib/compiler.js
```diff
@@ -513,16 +514,8 @@ function buildComponentInstructions(rootDir, componentConfig) {
...
-      if (name.indexOf('./') === 0 || name.indexOf('../') === 0) {
-        // Relative path
-        sourceFile = path.resolve(rootDir, name);
-      } else {
-        sourceFile = require.resolve(name);
-      }
-
       return {
-        sourceFile: sourceFile,
+        sourceFile: resolveAppScriptPath(rootDir, name),
```
bajtos: Before this change, a component path "mymod/file.js" was never resolved as relative, i.e. "{appRootDir}/mymod/file.js".

Is there a reasonably easy way how to preserve this behaviour? For example, can we add an `options` argument to `resolveAppScriptPath` that will be passed all the way down to `tryResolveAppPath` and which will control whether values that don't start with "./" or "../" are resolved as relative or not?

Example usage here:

```js
sourceFile: resolveAppScriptPath(rootDir, name, { strict: true })
```

Thoughts?

lib/compiler.js
```diff
@@ -538,3 +531,43 @@ function resolveRelativePaths(relativePaths, appRootDir) {
...
+    if (stats.isFile()) {
+      var otherFileExtension = path.extname(f);
+      if (otherFileExtension != FILE_EXTENSION_JSON &&
+        path.basename(f, otherFileExtension) == basename) {
+        results.push(otherFile);
+      }
+    }
+  });
+  return (results.length > 0 ? results[0] : filepath);
+}
```
bajtos: This code is very similar to what we have in [loadModuleDefinition()](https://github.com/PradnyaBaviskar/loopback-boot/blob/8e9c43c16b029ee534123bf2e0b37145c335431e/lib/compiler.js#L358-L385). Can you please refactor these two methods to share as much code as possible?

I also think the approach in `loadModuleDefinition` is better because it calls `tryReadDir` only once, while `fixFileExtension` is calling it for every file being resolved.

test/compiler.test.js
```diff
@@ -662,6 +688,75 @@ describe('compiler', function() {
...
+      var instructions = boot.compile({
+        appRootDir: appdir.PATH,
+        modelSources: ['./custom-models']
+      });
+
+      expect(instructions.models).to.have.length(1);
+      expect(instructions.models[0].sourceFile).to.equal(appJS);
+    });
+
+    it('resolves non-relative path in `modelSources` option', function() {
```
bajtos: Please rename to "resolves module relative path..." to keep this consistent with other test names.

test/compiler.test.js
```diff
@@ -662,6 +688,75 @@ describe('compiler', function() {
...
+      appdir.writeConfigFileSync('custom-models/car.json', { name: 'Car' });
+      var appJS = appdir.writeFileSync('custom-models/car.js', '');
+
+      var instructions = boot.compile(appdir.PATH);
+
+      expect(instructions.models).to.have.length(1);
+      expect(instructions.models[0].sourceFile).to.equal(appJS);
+    });
+
+    it('resolves non-relative path in `sources` option in `model-config.json`',
```
bajtos: Please rename to "resolves module relative path..." to keep this consistent with other test names.

test/compiler.test.js
```diff
@@ -1096,6 +1191,55 @@ describe('compiler', function() {
...
+      });
+
+      var instructions = boot.compile(appdir.PATH);
+
+      expect(instructions.middleware.middleware).to.have.length(1);
+      expect(instructions.middleware.middleware[0]).have.property(
+        'sourceFile', coffee);
+    });
+
+    it('prefers coffeescript over json for non-relative path middleware',
```
bajtos: Please rename to "... for module relative middleware path" to keep this consistent with other test names.

test/compiler.test.js
```diff
@@ -1195,6 +1339,63 @@ describe('compiler', function() {
...
+      expect(instructions.components[0]).have.property(
+        'sourceFile', appJS
+      );
+    });
+
+    it('loads component relative to node modules', function() {
+      appdir.writeConfigFileSync('component-config.json', {
+        'index': { }
+      });
+      var appJS = appdir.writeConfigFileSync('node_modules/index.js', '');
```
bajtos: It is rather unusual to have `node_modules/index.js`, typically the script is nested in a module directory, e.g. `node_modules/mycomponent/index.js`. Please rework the test to use `mycomponent` as the key in the config file and check that is is resolved to `node_modules/mycomponent/index.js`.

Please make the same change in the coffescript test below too.

test/compiler.test.js
```diff
@@ -1195,6 +1339,63 @@ describe('compiler', function() {
...
+      appdir.writeFileSync('component.json', '');
+
+      var instructions = boot.compile(appdir.PATH);
+
+      expect(instructions.components).to.have.length(1);
+      expect(instructions.components[0]).have.property(
+        'sourceFile', coffee);
+    });
+
+    it('prefers coffeescript over json for non-relative path component',
```
bajtos: Please rename to "... for module relative component path" to keep this consistent with other test names.

bajtos:
Reviewed, the patch looks much better now. I'd like you to refactor and simplify the implementation even more, see my comments above.

PradnyaBaviskar:
@bajtos, 
Updated the code, mainly to - 
 1. Refactor ```loadModuleDefinition``` and ```fixFileExtension``` functions.
 2. And secondly, added ```resolveOptions``` parameter to ```tryResolveAppPath()``` function, so as to retain backward compatibility.
 3. Also rephrased a few test cases, as suggested above.

Can you please review?

lib/compiler.js
```diff
@@ -538,3 +525,61 @@ function resolveRelativePaths(relativePaths, appRootDir) {
...
+  disallowedExtensions.forEach(function(ext) {
+    excludeExtensions[ext] = ext;
+  });
+
+  var ext = path.extname(filename);
+  return (ext in includeExtensions) && !(ext in excludeExtensions);
+}
+
+function fixFileExtension (filepath, files, disallowedExtensions,
+  contentDataType) {
```
PradnyaBaviskar: Will it be a good idea, to combine `disallowedExtensions` and `contentDataType` parameters, into a single parameter like `fixOptions`. An example for `fixOptions` would be
```
{
  'disallowedExtensions': ['.node'],
  'contentDataType': 'functions' 
}
```

lib/compiler.js
```diff
@@ -309,9 +310,17 @@ function resolveAppPath(rootDir, relativePath) {
...
+   * component path, `resolveOptions` parameter is added where
+   * `resolveOptions.strict` = true,
+   * means retain backward compatibility when resolving the path and
+   * `resolveOptions.strict` = false,
+   * does not enforce any such restriction when resolving the path */
+  resolveOptions = resolveOptions || { strict: false };
+
   var fullPath = path.resolve(rootDir, relativePath);
-  if (fs.existsSync(fullPath))
+  if (fs.existsSync(fullPath) && !resolveOptions.strict)
```
bajtos: I think this is not correct, as you will reject absolute paths in strict mode. Perhaps the following implementation is more correct?

```js
 if (name[0] === '/') {
  fullPath = relativePath;
} else if (name.indexOf('./') === 0 || name.indexOf('../') === 0 || !resolveOptions.strict) {
  fullPath = path.resolve(rootDir, relativePath);
}

if (fullPath && fs.existsSync(fullPath)) {
  return fullPath;
}

// resolve as a module now
```

lib/compiler.js
bajtos: > a single parameter like fixOptions

Sounds good, I think the name `options` is good enough.

Few more thoughts: `.node` should be ignored everywhere we ignore `.json`, it's just another extensions that gets special treatment from Node.js.

If find `contentDataType: 'function'` difficult to understand in the context of the calling code. Perhaps `onlyScriptsExportingFunction: true` would be a better option name?

PradnyaBaviskar:
@bajtos,
Done with the suggested changes! 

Please review.

bajtos:
@slnode test please

bajtos:
Landed, thank you!

PR #115, opened 2015-03-24
add console.error message to a bad require in a boot script

clarkbw:
I just hit this last night and thought it would be nice to have an error print out.  

I also found strongloop/loopback#1170 which I think this would fix but I don't agree that I should crash the server on a failing boot script, the error log should be enough to realize what happened.

The change produces the following error message in the tests:
`Failed loading boot script badScript.js`
`Cannot find module 'doesnt-exist’`

Let me know if you'd like to see any other changes.

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -231,13 +231,18 @@ function runScripts(app, list, callback) {
   var functions = [];
   list.forEach(function(filepath) {
     debug('Requiring script %s', filepath);
-    var exports = require(filepath);
-    if (typeof exports === 'function') {
-      debug('Exported function detected %s', filepath);
-      functions.push({
-        path: filepath,
-        func: exports
-      });
+    try {
+      var exports = require(filepath);
+      if (typeof exports === 'function') {
+        debug('Exported function detected %s', filepath);
+        functions.push({
+          path: filepath,
+          func: exports
+        });
+      }
+    } catch (err) {
+      console.error('Failed loading boot script: %s\n%s', filepath, err.stack);
+      throw err;
     }
   });
 


slnode:

rmg:
@slnode ok to test

rmg:
@clarkbw I'm trying to simplify the CLA auth, and to that end I've deleted the second app entirely from GitHub. Would you be able to try signing in _again_?

clarkbw:
> @clarkbw I'm trying to simplify the CLA auth, and to that end I've deleted the second app entirely from GitHub. Would you be able to try signing in again?

Sorry, still not working for me.  I revoked all SL auths.  I even signed out of Github and back in again.  I do have 2 factor auth turned on for GH but I doubt that affects things.

rmg:
@clarkbw ok, I've engaged the nuclear option and cleared everything.. one last try?

clarkbw:
> one last try

:fireworks: :fireworks: :fireworks: :fireworks: :fireworks:

rmg:
**FINALLY**

  ┬─┬ノ( º _ ºノ)

clarkbw:
updated and using `path.basename` instead of split tricks.

lib/executor.js
```diff
@@ -230,13 +230,19 @@ function runScripts(app, list, callback) {
...
+      if (typeof exports === 'function') {
+        debug('Exported function detected %s', filepath);
+        functions.push({
+          path: filepath,
+          func: exports
+        });
+      }
+    } catch (err) {
+      console.error('Failed loading boot script: %s\n%s',
+        path.basename(filepath), err.message);
```
bajtos: Why `path.basename` instead of the full path?

`err.message` is discarding the stack trace which usually contain a lot of useful information (IMO).

I am proposing the following:

```
console.error('Failed to load the boot script %s\n%s', filepath, err.stack)
```

Thoughts?

bajtos:
Thank you for the pull request @clarkbw, I left on comment to address - see above.

lib/executor.js
clarkbw: > Why path.basename instead of the full path?

It just made the output easier to read but if you had duplicate script names in different directories it could be a problem.

Your solution looks good, I'll update with that.

bajtos:

clarkbw:
:dancers:

PR #118, opened 2015-04-07
Use filename as default value for Model name

### LINKED ISSUE #42: Use filename as the default value for Model name
Use the file name (e.g. `product` for `common/models/product.json`) a default value that can be overridden by the name specified in the file.

If we process the file name using `_s.capitalize(_s.camelize(name))` where `_s` is the `underscore.string` module, both "order-item" and "OrderItem" will produce the same model name "OrderItem".

PradnyaBaviskar:
In reference to https://github.com/strongloop/loopback-boot/issues/42, made changes to use file name as the default value for Model name.

Close #42

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -7,6 +7,7 @@ var ConfigLoader = require('./config-loader');
 var debug = require('debug')('loopback:boot:compiler');
 var Module = require('module');
 var _ = require('lodash');
+var _s = require('underscore.string');
 
 var FILE_EXTENSION_JSON = '.json';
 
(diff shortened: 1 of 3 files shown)


bajtos:
Great job!

Could you please add few more tests similar to "uses file name as default value for model name" which will verify that different file name styles are handled as intented?

Examples to check: 
 - `order-item.json` -> OrderItem
 - `order_item.json` -> OrderItem
 - `order item.json` -> OrderItem

PradnyaBaviskar:
@bajtos -
Added test cases to verify...
> - order-item.json -> OrderItem
> - order_item.json -> OrderItem
> - order item.json -> OrderItem

Can you review?

bajtos:
Landed, thank you!

PR #119, opened 2015-04-10
upgrade lodash and drop underscore.string

clarkbw:
If we update to the latest stable lodash we can use the sting functions included and drop the usage of underscore.string.
`isFinite` changed from the older version of lodash to no longer coerce strings into numbers so we need to use `parseInt` before checking. `parseInt` on an `undefined` returns `NaN` which `isFinite` does not identify as a number.
Existing tests run checks for `flying-car` definitions to be named `FlyingCar` so there doesn’t seem to be a need to add more checks.

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -116,7 +116,9 @@ function setPort(app, instructions) {
     process.env.npm_package_config_port,
     app.get('port'),
     3000
-  ], _.isFinite);
+  ], function(v) {
+    return _.isFinite(parseInt(v, 10));
+  });
 
   if (port !== undefined) {
     var portType = typeof port;
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -1,5 +1,5 @@
 var assert = require('assert');
-var _ = require('lodash');
+var find = require('lodash/collection/find');
 var semver = require('semver');
 var debug = require('debug')('loopback:boot:executor');
 var async = require('async');
@@ -107,7 +107,7 @@ function setHost(app, instructions) {
 
 function setPort(app, instructions) {
   // jscs:disable requireCamelCaseOrUpperCaseIdentifiers
-  var port = _.find([
+  var port = find([
     process.env.npm_config_port,
     process.env.OPENSHIFT_SLS_PORT,
     process.env.OPENSHIFT_NODEJS_PORT,
@@ -116,9 +116,7 @@ function setPort(app, instructions) {
     process.env.npm_package_config_port,
     app.get('port'),
     3000
-  ], function(v) {
-    return _.isFinite(parseInt(v, 10));
-  });
+  ], function isNumberLike(v) { return Number.isFinite(parseInt(v, 10)); });
 
   if (port !== undefined) {
     var portType = typeof port;
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -6,7 +6,9 @@ var toposort = require('toposort');
 var ConfigLoader = require('./config-loader');
 var debug = require('debug')('loopback:boot:compiler');
 var Module = require('module');
-var _ = require('lodash');
+var uniq = require('lodash/array/uniq');
+var capitalize = require('lodash/string/capitalize');
+var camelCase = require('lodash/string/camelCase');
 
 var FILE_EXTENSION_JSON = '.json';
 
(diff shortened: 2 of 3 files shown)


slnode:

lib/executor.js
```diff
@@ -116,7 +116,9 @@ function setPort(app, instructions) {
     process.env.npm_package_config_port,
     app.get('port'),
     3000
-  ], _.isFinite);
+  ], function(v) {
+    return _.isFinite(parseInt(v, 10));
+  });
```
bajtos: Why is this change necessary? If I am reading the docs correctly, then lodash isFinite supports string values too. Note that `parseInt('10asd', 10)` return `10`, which is probably not what we want.

bajtos:
@slnode ok to test

lib/executor.js
bajtos: > Note that parseInt('10asd', 10) return 10, which is probably not what we want.

I take that back, the purpose of the port-scanning code is to find the first value that is defined.

In order to keep the size of the the browser bundle small, we are trying to get rid of lodash/underscore in as many places as possible. If `_.isFinite` in the new version of lodash does not work in the same way as the old one, then I would much rather rewrite this condition so that it does not depend on lodash at all.

I think the following may serve pretty well:

```js
function isNumberLike(v) { return Number.isFinite(parseInt(v, 10)); }
```

bajtos:
@clarkbw thank you for the pull request, I am all for making the number of dependencies smaller by removing `underscore.string`. Can you please address my comment above?

lib/executor.js
clarkbw: > In order to keep the size of the the browser bundle small, we are trying to get rid of lodash/underscore in as many places as possible

I think that's a worthwhile goal.  However some functions like the String features of `_.capitalize` and `_.camelCase` are IMO worth importing from something like lodash instead of having our own copies which could receive little testing compared to the lodash versions.

To help with the browserify sizing we can move to a pattern like this:
```
var capitalize = require('lodash/string/capitalize');
var camelCase = require('lodash/string/camelCase');
```
Which would allow browserify to only pull in the required functions instead of all of lodash and would make it clearer why this module requires lodash at all.  

@bajtos What do you think of that?

clarkbw: > I think the following may serve pretty well:

re: your function, it does look like a good replacement for the lodash function.  I'll update the patch to use that and try a version with my suggestion from above.

clarkbw:
@bajtos let me know how this updated patch looks

bajtos:
@clarkbw I have improved your patch a little bit and landed via d9c0f12:
 - Executor no longer depends on lodash, the find method can be easily rewritten via array.filter.
 - It's ok to use lodash in the compiler, that part should not be included in the browser bundle.
 - I have reworded the commit message a bit.

I hope you don't mind that I took this over, I figured out it would be easier to make the changes myself rather than explaining them and asking you to make them.

clarkbw:
> I hope you don't mind that I took this over, I figured out it would be easier to make the changes myself rather than explaining them and asking you to make them.

Not a problem, I don't mind you taking over at all.  

I am surprised that you didn't want to go the individual import route.  Would you mind explaining why you chose to import all of lodash vs. the model I used in 5fc271e7af9badea732a6474699afbecad0402a0 where I did an import like `var capitalize = require('lodash/string/capitalize');` instead.  With the browserify model in mind I felt like this was the best path forward and recommended from [lodash](https://github.com/lodash/lodash/tree/3.0.0-npm#installation).  This is pure curiosity and not questioning your decision.  Thank you!

bajtos:
> I am surprised that you didn't want to go the individual import route. Would you mind explaining why you chose to import all of lodash vs. the model I used in 5fc271e where I did an import like var capitalize = require('lodash/string/capitalize'); instead. With the browserify model in mind I felt like this was the best path forward and recommended from lodash. This is pure curiosity and not questioning your decision. Thank you!

Sure. The compiler (lib/compiler.js) is not included in the browser bundle created by browserify (see [package.json#L16](https://github.com/strongloop/loopback-boot/blob/6c0c8b9f8b3cf371b4d5ed847e4b014299878642/package.json#L16) and [browser.js](https://github.com/strongloop/loopback-boot/blob/6c0c8b9f8b3cf371b4d5ed847e4b014299878642/browser.js)), therefore it does not matter how much of lodash we import there.

In which case I prefer the solution that involves less code changes, as unnecessary code changes make code archaeology more difficult.

Does it make sense? Please let me know if there is anything to explain in more details.

clarkbw:
That's what I was wondering, makes perfect sense then.  Thanks!

PR #120, opened 2015-04-14
Add support for mixins

PradnyaBaviskar:
In context of - https://github.com/strongloop/loopback-boot/issues/79, added support for mixins.

@bajtos, can you please review?
This PL is halfway done - I'm still working on 
 1. Adding test cases for the executor.
 2. Creating a common function, that can be used by both ```findModelDefinitions``` and ```buildAllMixinInstructions```.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -84,6 +84,10 @@ module.exports = function compile(options) {
   var modelInstructions = buildAllModelInstructions(
     modelsRootDir, modelsConfig, modelSources);
 
+  var mixinDirs = options.mixinDirs || [];
+  var mixinInstructions = buildAllMixinInstructions(
+    appRootDir, mixinDirs, options);
+
   // When executor passes the instruction to loopback methods,
   // loopback modifies the data. Since we are loading the data using `require`,
   // such change affects also code that calls `require` for the same file.
@@ -93,6 +97,7 @@ module.exports = function compile(options) {
     models: modelInstructions,
     middleware: middlewareInstructions,
     components: componentInstructions,
+    mixins: mixinInstructions,
     files: {
       boot: bootScripts
     }
@@ -135,10 +140,11 @@ function assertIsValidModelConfig(config) {
  * @private
  */
 
-function findScripts(dir) {
+function findScripts(dir, extensions) {
   assert(dir, 'cannot require directory contents without directory name');
 
   var files = tryReadDir(dir);
+  extensions = extensions || _.keys(require.extensions);
 
   // sort files in lowercase alpha for linux
   files.sort(function(a, b) {
@@ -598,4 +604,70 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
   var files = tryReadDir(sourceDir);
   var fixedFile = fixFileExtension(resolvedPath, files, false);
   return (fixedFile === undefined ? resolvedPath : fixedFile);
+}
+
+function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
+  var extensions = _.without(_.keys(require.extensions),
+    _.keys(getExcludedExtensions()));
+  var files = options.mixins || [];
+
+  mixinDirs.forEach(function(dir) {
+    dir = tryResolveAppPath(appRootDir, dir);
+    if (!dir) {
+      debug('Skipping unknown module source dir %j', dir);
+      return;
+    }
+    files = files.concat(findScripts(dir, extensions));
+  });
+
+  var mixins = files.map(function(filepath) {
+    var dir = path.dirname(filepath);
+    var ext = path.extname(filepath);
+    var name = path.basename(filepath, ext);
+    var metafile = path.join(dir, name + FILE_EXTENSION_JSON);
+
+    name = normalizeMixinName(name, options);
+    var meta = {};
+    meta.name = name;
+    if (fs.existsSync(metafile)) {
+      // May overwrite name, not sourceFile
+      _.extend(meta, require(metafile));
+    }
+    meta.sourceFile = filepath;
+    return meta;
+  });
+
+  return mixins;
+}
+
+function normalizeMixinName(str, options) {
+  var normalization = options.normalization;
+  switch (normalization) {
+    case false:
+    case 'none': return str;
+
+    case undefined:
+    case 'classify':
+      str = String(str).replace(/([A-Z]+)/g, ' $1').trim();
+      str = String(str).replace(/[\W_]/g, ' ').toLowerCase();
+      str = str.replace(/(?:^|\s|-)\S/g, function(c) { return c.toUpperCase(); });
+      str = str.replace(/\s+/g, '');
+      return str;
+
+    case 'dasherize':
+      str = String(str).replace(/([A-Z]+)/g, ' $1').trim();
+      str = String(str).replace(/[\W_]/g, ' ').toLowerCase();
+      str = str.replace(/\s+/g, '-');
+      return str;
+
+    default:
+      if (typeof normalization === 'function') {
+        return normalization(str);
+      }
+
+      var err = new Error('Invalid normalization format - "' +
+        normalization + '"');
+      err.code = 'INVALID_NORMALIZATION_FORMAT';
+      throw err;
+  }
 }
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -160,6 +160,7 @@ function setupDataSources(app, instructions) {
 }
 
 function setupModels(app, instructions) {
+  defineMixins(app, instructions);
   defineModels(app, instructions);
 
   instructions.models.forEach(function(data) {
@@ -167,6 +168,26 @@ function setupModels(app, instructions) {
     if (!data.config) return;
 
     app.model(data._model, data.config);
+  });
+}
+
+function defineMixins(app, instructions) {
+  var modelBuilder = (app.registry || app.loopback).modelBuilder;
+  var BaseClass = app.loopback.Model;
+  var mixins = instructions.mixins || [];
+
+  if (!modelBuilder.mixins || !mixins.length) return;
+
+  mixins.forEach(function(obj) {
+    var mixin = require(obj.sourceFile);
+
+    if (typeof mixin === 'function' || mixin.prototype instanceof BaseClass) {
+      debug('Defining mixin %s', obj.name);
+      modelBuilder.mixins.define(obj.name, mixin); // TODO (name, mixin, meta)
+    } else {
+      debug('Skipping mixin file %s - `module.exports` is not a function' +
+        ' or Loopback model', obj);
+    }
   });
 }
 
--- a/test/browser.test.js
+++ b/test/browser.test.js
@@ -61,6 +61,27 @@ describe('browser support', function() {
     });
   });
 
+  it('loads mixins', function(done) {
+    var appDir = path.resolve(__dirname, './fixtures/browser-app');
+    var options = {
+      appRootDir: appDir,
+      mixinDirs: ['./mixins']
+    };
+
+    browserifyTestApp(options, function(err, bundlePath) {
+      if (err) return done(err);
+
+      var app = executeBundledApp(bundlePath);
+
+      var modelBuilder = app.registry.modelBuilder;
+      var registry = modelBuilder.mixins.mixins;
+      expect(Object.keys(registry)).to.eql(['TimeStamps']);
+      expect(app.models.Customer.timeStampsMixin).to.eql(true);
+
+      done();
+    });
+  });
+
   it('supports coffee-script files', function(done) {
     // add coffee-script to require.extensions
     require('coffee-script/register');
@@ -82,7 +103,7 @@ describe('browser support', function() {
   });
 });
 
-function browserifyTestApp(appDir, strategy, next) {
+function browserifyTestApp(options, strategy, next) {
   // set default args
   if (((typeof strategy) === 'function') && !next) {
     next = strategy;
@@ -91,9 +112,10 @@ function browserifyTestApp(appDir, strategy, next) {
   if (!strategy)
     strategy = 'default';
 
+  var appDir = typeof(options) === 'object' ? options.appRootDir : options;
   var b = compileStrategies[strategy](appDir);
 
-  boot.compileToBrowserify(appDir, b);
+  boot.compileToBrowserify(options, b);
 
   exportBrowserifyToFile(b, 'browser-app-bundle.js', next);
 }
--- a/test/compiler.test.js
+++ b/test/compiler.test.js
@@ -1014,6 +1014,160 @@ describe('compiler', function() {
       instructions = boot.compile(appdir.PATH);
       expect(instructions.config).to.not.have.property('modified');
     });
+
+    describe('for mixins', function() {
+      function verifyMixinIsFoundViaMixinDirs(sourceFile, mixinDirs) {
+        var appJS = appdir.writeFileSync(sourceFile, '');
+
+        var instructions = boot.compile({
+          appRootDir: appdir.PATH,
+          mixinDirs: mixinDirs
+        });
+
+        expect(instructions.mixins[0].sourceFile).to.eql(appJS);
+      }
+
+      it('supports `mixinDirs` option', function() {
+        verifyMixinIsFoundViaMixinDirs('mixins/other.js', ['./mixins']);
+      });
+
+      it('resolves relative path in `mixinDirs` option', function() {
+        verifyMixinIsFoundViaMixinDirs('custom-mixins/vehicle.js',
+          ['./custom-mixins']);
+      });
+
+      it('resolves module relative path in `mixinDirs` option', function() {
+        verifyMixinIsFoundViaMixinDirs('node_modules/custom-mixins/vehicle.js',
+          ['custom-mixins']);
+      });
+
+      describe('name normalization', function() {
+        var options;
+        beforeEach(function() {
+          options = { appRootDir: appdir.PATH, mixinDirs: ['./mixins'] };
+
+          appdir.writeFileSync('mixins/foo.js', '');
+          appdir.writeFileSync('mixins/time-stamps.js', '');
+          appdir.writeFileSync('mixins/camelCase.js', '');
+          appdir.writeFileSync('mixins/PascalCase.js', '');
+          appdir.writeFileSync('mixins/space name.js', '');
+        });
+
+        it('supports classify', function() {
+          options.normalization = 'classify';
+          var instructions = boot.compile(options);
+
+          var mixins = instructions.mixins;
+          var mixinNames = mixins.map(getNameProperty);
+
+          expect(mixinNames).to.eql([
+            'CamelCase', 'Foo', 'PascalCase', 'SpaceName', 'TimeStamps'
+          ]);
+        });
+
+        it('supports dasherize', function() {
+          options.normalization = 'dasherize';
+          var instructions = boot.compile(options);
+
+          var mixins = instructions.mixins;
+          var mixinNames = mixins.map(getNameProperty);
+
+          expect(mixinNames).to.eql([
+            'camel-case', 'foo', 'pascal-case', 'space-name', 'time-stamps'
+          ]);
+        });
+
+        it('supports custom function', function() {
+          var normalize = function(name) { return name.toUpperCase(); };
+          options.normalization = normalize;
+          var instructions = boot.compile(options);
+
+          var mixins = instructions.mixins;
+          var mixinNames = mixins.map(getNameProperty);
+
+          expect(mixinNames).to.eql([
+            'CAMELCASE', 'FOO', 'PASCALCASE', 'SPACE NAME', 'TIME-STAMPS'
+          ]);
+        });
+
+        it('supports none', function() {
+          options.normalization = 'none';
+          var instructions = boot.compile(options);
+
+          var mixins = instructions.mixins;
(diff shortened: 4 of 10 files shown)


PradnyaBaviskar:
Added test case for executor.

lib/compiler.js
```diff
@@ -136,10 +141,11 @@ function assertIsValidModelConfig(config) {
  * @private
  */
 
-function findScripts(dir) {
+function findScripts(dir, extensions) {
   assert(dir, 'cannot require directory contents without directory name');
 
   var files = tryReadDir(dir);
+  extensions = [].concat(extensions || _.keys(require.extensions));
```
bajtos: Hmm, is this the same as the following simpler form? `extensions = extensions || _.keys(require.extensions)`?

lib/compiler.js
```diff
@@ -598,3 +604,67 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
...
+    meta.name = name;
+    if (fs.existsSync(metafile)) {
+      // May overwrite name, not id or filepath
+      _.extend(meta, require(metafile));
+    }
+    meta.filepath = filepath;
+    mixins[meta.name] = meta;
+  });
+
+  return mixins;
```
bajtos: The instructions entity should be an array of mixins to create, not a lookup table. That is the convention we use elsewhere, see model instructions for an example.

test/compiler.test.js
```diff
@@ -1014,6 +1014,201 @@ describe('compiler', function() {
...
+
+      it('supports mixin name normalization - dasherize', function() {
+        appdir.writeFileSync('mixins/foo.js', '');
+        appdir.writeFileSync('mixins/time-stamps.js', '');
+
+        var options = { appRootDir: appdir.PATH, normalization: 'dasherize' };
+        var instructions = boot.compile(options);
+
+        expect(instructions.mixins).to.have.property('foo');
+        expect(instructions.mixins).to.have.property('time-stamps');
```
bajtos: Please check the following two file names too,  ideally in all "supports mixing name normalization" tests:

 - `mixins/camelCase.js`
 - `mixins/PascalCase.js`

test/compiler.test.js
```diff
@@ -1014,6 +1014,201 @@ describe('compiler', function() {
...
+        var instructions = boot.compile({
+          appRootDir: appdir.PATH,
+          mixinSources: ['custom-mixins']
+        });
+
+        expect(instructions.mixins).to.have.property('Other');
+        expect(instructions.mixins).to.not.have.property('Default');
+      });
+
+      it('supports mixin name normalization - classify', function() {
```
bajtos: Please wrap all tests called "supports mixin name normalization" in a new "describe" block and add a beforeEach hook to create the mixin files. 

We should probably test file names containing a space too.

```js
describe('name normalization', function() {
  beforeEach(function() {
    appdir.writeFileSync('mixins/foo.js', '');
    appdir.writeFileSync('mixins/time-stamps.js', '');
    appdir.writeFileSync('mixins/camelCase.js', '');
    appdir.writeFileSync('mixins/PascalCase.js', '');
    appdir.writeFileSync('mixins/space name.js', '');
  });

  it('supports "classify"', function() {
    var instructions = //;
    expect(_.keys(instructions.mixins)).to.eql([
      'Foo', 'TimeStamps', 'CamelCase', 'PascalCase', 'SpaceName'
    ]);
  });
  // etc.
});
```

test/executor.test.js
```diff
@@ -271,6 +271,35 @@ describe('executor', function() {
...
+        var modelBuilder = app.loopback.modelBuilder;
+        var meta = modelBuilder.mixins.meta;
+        var json = require('./fixtures/simple-app/mixins/time-stamps.json');
+
+        expect(meta.Timestamping.name).to.equal(json.name);
+        expect(meta.Timestamping).to.have.property('filepath');
+        expect(meta.Timestamping.description).to.equal(json.description);
+        expect(meta.Timestamping.options).to.eql(json.options);
+      });
+    });
```
bajtos: When looking back, writing unit-tests via `simpleAppInstructions()` was a mistake. Please write the new tests using the approach where all files are created via `appdir.writeFileSync` and friends, as you already do in compiler tests. Other executor tests are already following this pattern, see for example ["it disables component when configuration is not set"](https://github.com/PradnyaBaviskar/loopback-boot/blob/9704c70f465947465d9bf5e5976a036cebfae072/test/executor.test.js#L506-L519).

test/compiler.test.js
```diff
@@ -1014,6 +1014,201 @@ describe('compiler', function() {
...
+        var options = { appRootDir: appdir.PATH, normalization: 'classify' };
+        var instructions = boot.compile(options);
+
+        expect(instructions.mixins).to.eql({
+          FooBar: {
+            name: 'FooBar',
+            description: 'JSON file name same as JS file name',
+            filepath: appJS
+          }
+        });
```
bajtos: I am concerned about mixing metadata from the JSON file with boot data like `filepath`. I would rather use the same approach that we use for models, see [this code](https://github.com/PradnyaBaviskar/loopback-boot/blob/1b9f13706229a80877012428fceeda6dbea85356/lib/compiler.js#L203-L208). Also `filepath` should be called `sourceFile`.

bajtos:
The code looks good in general, I left few comments to address.

bajtos:
> Creating a common function, that can be used by both findModelDefinitions and buildAllMixinInstructions.

Don't worry about this part too much. If there isn't an easy solution then the current implementation is relatively ok. Model definitions are searching for JSON files, while Mixins are searching of non-JSON files, it makes sense that the implementation is slightly different.

bajtos:
@PradnyaBaviskar what's your opinion on https://github.com/strongloop/loopback-boot/pull/111#discussion_r27220710? Can you include the necessary changes in this pull request? Alternatively you can rename `mixinSources` to `mixinDirs` and implement `mixinSources` later in a new pull request. Let me know what would work best for you.

PradnyaBaviskar:
@bajtos,
Addressed all the comments suggested by you, above. 
Please review.

PradnyaBaviskar:
>@PradnyaBaviskar what's your opinion on #111 (comment)?

I agree, ```mixinSources``` must be renamed to ```mixinDirs```, to be consistent with previous work. I've   worked on this change in the current PL. And, will plan to submit a separate PL for  implementing ```mixinSources```

bajtos:
@PradnyaBaviskar many executor tests are failing, could you please run `npm test` locally and fix the failures?

This is what I see on our build server:

```
TypeError: Cannot read property 'mixins' of undefined
    at defineMixins (/mnt/jenkins/workspace/loopback-boot.pr/eb78835e/lib/executor.js:180:19)
    at setupModels (/mnt/jenkins/workspace/loopback-boot.pr/eb78835e/lib/executor.js:165:3)
    ...
```

BTW I have enabled Travis CI for this repository. The next time you push a change, Travis CI should build it and add a link to the "checks" panel where you can see the build result.

test/executor.test.js
```diff
@@ -271,6 +271,62 @@ describe('executor', function() {
         done();
       });
     });
+
+    describe ('for mixins', function() {
+      it('defines `mixins` files', function() {
+        var example = appdir.writeFileSync('mixins/example.js',
+          'module.exports = ' +
+          'function(app, options) { app.mixinExample = true; }');
```
bajtos: Since the mixin is never called, you can export an empty function. Also note that the first function arg is `Model`, not `app`.

```js
module.exports = function(Model, options) {};
```

Please apply the same change for `mixins/time-stamps.js` too.

lib/compiler.js
```diff
@@ -85,6 +85,10 @@ module.exports = function compile(options) {
   var modelInstructions = buildAllModelInstructions(
     modelsRootDir, modelsConfig, modelSources);
 
+  var mixinDirs = options.mixinDirs || modelsMeta.mixins || ['./mixins'];
```
bajtos: `mixinDirs` should be empty by default. The value `['./mixins']` should be used as default for `mixinSources` only.

test/executor.test.js
```diff
@@ -271,6 +271,62 @@ describe('executor', function() {
         done();
       });
     });
+
+    describe ('for mixins', function() {
+      it('defines `mixins` files', function() {
```
bajtos: I find this test name confusing. How about "defines mixins from instructions"?

test/executor.test.js
```diff
@@ -271,6 +271,62 @@ describe('executor', function() {
...
+
+        var registry = modelBuilder.mixins.mixins;
+        expect(Object.keys(registry)).to.eql(['Example', 'Timestamping']);
+
+        var meta = modelBuilder.mixins.meta;
+        var sourceFiles = meta.map(getSourceFileProperty);
+        expect(sourceFiles).to.eql([example, ts]);
+      });
+
+      it('supports mixin as model', function() {
```
bajtos: I feel the test name does not describe the purpose of test well. How about "applies mixin metadata specified in instructions"?

test/fixtures/simple-app/mixins/example.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function(Model, options) {
+
+  Model.exampleMixin = true;
+
+};
```
bajtos: I think this file is no longer needed, please delete.

test/fixtures/simple-app/mixins/time-stamps.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function(Model, options) {
+
+  Model.timeStampsMixin = true;
+
+};
```
bajtos: I think this file is no longer needed, please delete.

test/fixtures/simple-app/mixins/time-stamps.json
```diff
@@ -0,0 +1,13 @@
...
+  "version": "0.0.1",
+  "description": "Track create and update",
+  "keywords": ["model", "date", "time"],
+  "options": {
+    "postfix": { 
+      "type": "string",
+      "description": "The postfix string",
+      "default": "At"
+    }
+  }
```
bajtos: I think this file is no longer needed, please delete.

bajtos:
@PradnyaBaviskar thank you for the update, I left a bunch of comments above.

There is one more major thing to implement before this PR can be landed: support for browserify.

 - Extend the fixture [browser-app](https://github.com/strongloop/loopback-boot/tree/master/test/fixtures/browser-app) and add a mixin file. Browser tests are not creating the test app files from within the test code yet.
 - Add a new test to [test/browser.test.js](https://github.com/strongloop/loopback-boot/blob/master/test/browser.test.js) which will configure `boot` with `mixinDirs` pointing to the mixin and check that the mixin was correctly loaded.
 - Implementation wise, look at [lib/bundler.js](https://github.com/strongloop/loopback-boot/blob/master/lib/bundler.js). Since mixin instructions use `sourceFile` property now, the implementation should be just another variant of existing `bundleModelScripts` and `bundleComponentScripts`.

lib/compiler.js
bajtos: Also `modelsMeta.mixins` should be used to initialise `mixinSources`, not `mixinDirs`.

```js
var mixinDirs = options.mixinDirs || [];
var mixinSources = options.mixinSources || modelsMeta.mixins || ['./mixins'];
```

PradnyaBaviskar:
@bajtos, done with the suggested changes. Please review.

lib/compiler.js
```diff
@@ -598,3 +604,69 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
...
+    var dir = path.dirname(filepath);
+    var ext = path.extname(filepath);
+    var name = path.basename(filepath, ext);
+    var metafile = path.join(dir, name + FILE_EXTENSION_JSON);
+
+    name = normalizeMixinName(name, options);
+    var meta = {};
+    meta.name = name;
+    if (fs.existsSync(metafile)) {
+      // May overwrite name, not id or filepath
```
bajtos: Please remove the mention of "id" from the comment, there is no id property now.

lib/executor.js
```diff
@@ -167,6 +168,28 @@ function setupModels(app, instructions) {
   });
 }
 
+function defineMixins(app, instructions) {
+  var modelBuilder = (app.registry || app.loopback).modelBuilder;
+  var BaseClass = app.loopback.Model;
+  var mixins = instructions.mixins || [];
+  if (modelBuilder.mixins && mixins.length > 0) {
```
bajtos: Use reverse logic + early return to reduce nesting level.

```js
if (!modelBuilder.mixins || !mixins.length) return;
```

lib/executor.js
```diff
@@ -167,6 +168,28 @@ function setupModels(app, instructions) {
...
+    modelBuilder.mixins.meta = [];
```
bajtos: Let's remove `mixins.meta` for now. AFAIK, it's not used anywhere. We can always introduce it back later.

lib/executor.js
```diff
@@ -167,6 +168,28 @@ function setupModels(app, instructions) {
...
+    mixins.forEach(function(obj) {
+      var meta = obj;
+      var mixin = require(obj.sourceFile);
+
+      if (typeof mixin === 'function' || mixin.prototype instanceof BaseClass) {
+        debug('Configuring mixin %s', obj.name);
```
bajtos: Please change to "Defining mixin %s".

test/browser.test.js
```diff
@@ -61,6 +61,32 @@ describe('browser support', function() {
...
+
+    browserifyTestApp(options, function(err, bundlePath) {
+      if (err) return done(err);
+
+      var app = executeBundledApp(bundlePath);
+
+      var modelBuilder = (app.registry || app.loopback).modelBuilder;
+
+      var meta = modelBuilder.mixins.meta;
+      var ts = require('./fixtures/browser-app/mixins/time-stamps.json');
```
bajtos: I am proposing to simplify this test to check that `modelBuilder.mixins.TimeStamps` is defined and is a function. That enables us to remove `mixins.meta` as I proposed above.

test/compiler.test.js
```diff
@@ -1014,6 +1014,176 @@ describe('compiler', function() {
       instructions = boot.compile(appdir.PATH);
       expect(instructions.config).to.not.have.property('modified');
     });
+
+    describe('for mixins', function() {
+      function verifyMixinsUsingOptions(sourceFile, mixinDirs) {
```
bajtos: Please use a more descriptive name, e.g. "verifyMixinIsFoundViaMixinDirs"

test/compiler.test.js
```diff
@@ -1014,6 +1014,176 @@ describe('compiler', function() {
...
+          appdir.writeFileSync('mixins/foo.js', '');
+          appdir.writeFileSync('mixins/time-stamps.js', '');
+          appdir.writeFileSync('mixins/camelCase.js', '');
+          appdir.writeFileSync('mixins/PascalCase.js', '');
+          appdir.writeFileSync('mixins/space name.js', '');
+        });
+
+        it('supports classify', function() {
+          var options = { appRootDir: appdir.PATH, normalization: 'classify',
+          mixinDirs: ['mixins'] };
```
bajtos: Please use a different indentation, I find this style difficult to read. It's also not clear which part is important and test-specific, and which part is generic.

Perhaps you can create `options` in `beforeEach` hook and then set only the normalization property in individual tests?

```js
// beforeEach
options = { appRootDir: appdir.PATH, mixinDirs: ['mixins'] }

// in the test
options.normalization = 'classify';
var instructions = boot.compile(options);
```

lib/compiler.js
```diff
@@ -598,3 +604,69 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
...
   return (fixedFile === undefined ? resolvedPath : fixedFile);
 }
+
+function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
+  var extensions = _.without(_.keys(require.extensions),
+    _.keys(getExcludedExtensions()));
+  var files = options.mixins || [];
+
+  mixinDirs.forEach(function(dir) {
+    dir = tryResolveAppPath(appRootDir, dir);
```
bajtos: Please use the strict mode here, so that `mixins` is always resolved via `node_modules` and never as `./mixins`. 

I think it will be best to make the `strict` option enabled by default, and explicitly set `strict:false` only in the few places where we need to preserve backwards compatibility.

Thoughts?

test/compiler.test.js
```diff
@@ -1014,6 +1014,176 @@ describe('compiler', function() {
...
+          var instructions = boot.compile(options);
+
+          var mixins = instructions.mixins;
+          var mixinNames = mixins.map(getNameProperty);
+
+          expect(mixinNames).to.eql(['CamelCase', 'Foo',
+            'PascalCase', 'SpaceName', 'TimeStamps']);
```
bajtos: Please use a different indentation, so that the list of items is not split in the middle. Few examples illustrating style that I find easier to read:

```js
expect(mixinNames).to.eql([
  'CamelCase', 'Foo', 'PascalCase', 'SpaceName', 'TimeStamps'
]);

expect(mixinNames)
  .to.eql(['CamelCase', 'Foo', 'PascalCase', 'SpaceName', 'TimeStamps']);
```

The same comment applies to all normalization tests below.

test/compiler.test.js
```diff
@@ -1014,6 +1014,176 @@ describe('compiler', function() {
...
+        });
+      });
+
+      it('throws error for invalid normalization format', function() {
```
bajtos: I think this should be nested under "name normalization", for example as "(it) throws error for invalid format"

test/executor.test.js
```diff
@@ -271,6 +271,67 @@ describe('executor', function() {
...
+              'default': 'At'
+            }
+          }
+        });
+
+        var options = { appRootDir: appdir.PATH,
+          mixinDirs: ['mixins'] };
+        boot(app, options);
+
+        var modelBuilder = (app.registry || app.loopback).modelBuilder;
```
bajtos: I think it's ok to rely on `app.registry` being present when inside a unit test. loopback-boot controls the version of loopback used for unit-tests via the devDependency in `package.json`.

Could you please do the following?

 - bump up the [loopback version in package.json](https://github.com/strongloop/loopback-boot/blob/e3763ac28e723d351a1423abc8e068916fbe49d7/package.json#L41) to "^2.16.3" 
 - use `app.registry.modelBuilder` in unit-tests (`test/*.js`)

Note that `lib/*.js` files must support older loopback versions, so they should not be changed.

test/fixtures/browser-app/mixins/time-stamps.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function(Model, options) {
+
+  Model.timeStampsMixin = true;
```
bajtos: AFAICT, the browser test is not checking whether `Model.timeStampsMixin` was set. Please modify the browser fixture app to use this mixin and then verify in the unit-test that the flag was set.

I think this is what you need to do:
 1. Modify the file `test/fixtures/browser-app/models/customer.json` and add `"mixins": {"TimeStamps": {} }`
 2. In the test, check that `app.models.Customer.timeStampsMixin` is `true`.

bajtos:
Reviewed, you did a good job! The patch is getting close to be landed. To make future review easier for me, please stop rewriting the history and send new changes in a new commit.

lib/compiler.js
PradnyaBaviskar: Per the comment in the code 
```
   * `resolveOptions.strict` = true,
   * means retain backward compatibility when resolving the path and
   * `resolveOptions.strict` = false,
   * does not enforce any such restriction when resolving the path
```
When you say
> explicitly set strict:false only in the few places where we need to preserve backwards compatibility.

I'm a bit confused, as ```strict:true``` retains backward compatibility.

bajtos: I guess the comment is not clear enough then.

For component paths (i.e. when resolving keys in `component-config.json`), the backwards-compatible behaviour is to always resolve "foo" as "node_modules/foo" and never as "./foo".

On the other hand, for some other places like model sources, the backwards-compatible behaviour is to resolve "foo" as "./foo" or "node_modules/foo".

The strict option controls this behaviour. Strict:true = search `node_modules` only. Strict:false = try a relative path too.

Now my argument is that all new code should use strict:true mode and therefore it's better to change `tryResolveAppPath` to default to strict:true.

I have made the cleanup myself, could you please review https://github.com/strongloop/loopback-boot/pull/127?

test/browser.test.js
```diff
@@ -61,6 +61,28 @@ describe('browser support', function() {
...
+    };
+
+    browserifyTestApp(options, function(err, bundlePath) {
+      if (err) return done(err);
+
+      var app = executeBundledApp(bundlePath);
+
+      var modelBuilder = app.registry.modelBuilder;
+      var registry = modelBuilder.mixins.mixins;
+      expect(Object.keys(registry)).to.eql(['Timestamping']);
```
bajtos: Inside `test/fixtures/browser-app/mixins/time-stamps.json`, you are renaming the mixin to `Timestamping`, but the model configuration refers to original name `TimeStamps`. 

I am proposing to delete `test/fixtures/browser-app/mixins/time-stamps.json` and keep the script file only.

PradnyaBaviskar:
@bajtos, 
I've pushed the latest changes. Can you review?

bajtos:
LGTM. Can you please squash the commits into a single one and provide a descriptive commit message explaining the new boot option "mixinDirs"?

bajtos:
Landed, thank you!

Issue #123, opened 2015-04-20
Unable to boot in the browser - Number.isFinite is not a function.


mrfelton:
The change at https://github.com/strongloop/loopback-boot/commit/044a4df07a8c076ec0c72e98261fdb44631f7acd#diff-3a73542dc68d28ea0d34b33d6df62f85R119 breaks when run in the browser via browserify.

When I'm running unit tests in PhantomJS using karma I get the following fatal error:

"TypeError: 'undefined' is not a function (evaluating 'Number.isFinite(parseInt(v, 10))')
at /devenv/build/client/lbclient/browser.bundle.js:1282"

Number.isFinite is not well supported in browsers.

mrfelton:
Patch above updates to use the polyfill code detailed on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite in place of Number.isFinite()

raymondfeng:
@mrfelton Can you submit your patch as a github pull request?

Amir-61:
[PR#126](https://github.com/strongloop/loopback-boot/pull/126) fixes the issue. closing the ticket now.

Thanks

PR #124, opened 2015-04-20
Use polyfill code for Number.isFinite

mrfelton:
Fix for #123

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -116,7 +116,7 @@ function setPort(app, instructions) {
     app.get('port'),
     3000
   ], function isNumberLike(v) {
-    return Number.isFinite(parseInt(v, 10));
+    return typeof v === "number" && isFinite(v);
   });
 
   if (port !== undefined) {


slnode:

bajtos:
@mrfelton thank you for the pull request. I am afraid it is a breaking change. Before we made modifications to this part, loopback-boot was accepting any value that could be parsed as an integer, e.g "10abc". (Unfortunately this was not covered by tests.)

Can you please rework the patch to honour this behaviour and also add a unit-test to make sure this behaviour is covered by our test suite?

bajtos:
@slnode ok to test

bajtos:
@mrfelton I also need you to sign our CLA before I can land your patch, see the failed "clahub" check.

bajtos:
Closing in favour or https://github.com/strongloop/loopback-boot/pull/126.

FWIW, the claim *loopback-boot was accepting any value that could be parsed as an integer, e.g "10abc"* was actually wrong.

Issue #82, opened 2014-12-10
Warn the user when a config file was not found, but overrides exist


bajtos:
See #71:

> A few weeks back I was reading loopback's confluence docs and came across the config page. I saw the alternative config options that loopback supports and I thought: "oh, so I could use a smarter config! Great I need that". so I went ahead, renamed config.json to config.local.js and started converting the code. Substituting json config with a js config was second nature because I've seen it in many projects including my own. I restarted the server and realized the config was not picked up. I went back to the docs page and still couldn't figure it out. It was only after reading the source code that I was able to understand the config requirements. it took a few hours overall as I couldn't immediately spot where configuration is processed.

We should detect such situation and print a warning.

Related: #80.

superkhau:
Connect to #162

Issue #80, opened 2014-12-09
Dynamic config options


bajtos:
It turns out that per-environment dynamic overrides via `config.{env}.js` are not good enough, as many users prefer to keep all config in a single javascript file (see e.g. #71). The current solution is to use `config.local.js` for that, even thought that file was initially intended for machine-local configuration that should not be checked into VCS (git, svn).

I am proposing to:
- [ ] Add a new suffix similar to `local`, files with this suffix will be always loaded. Possible suffixes: `dynamic`, `all`. This new suffix should be supported at least by `datasources.json`, `config.json` and `middleware.json`.
- [ ] Update the documentation and explain the difference between the new suffix and `local`, describe when to use which one.
- [ ] Add new unit tests to loopback-workspace to ensure it can handle missing config.json file. Possibly also change the implementation so that it does not create an empty config.json file.
- [ ]  Make the file `config.json` (and only this file) optional.

bajtos:
Related: https://github.com/strongloop/loopback-boot/issues/84

MarcDiethelm:
The current situation is surprisingly counter-intuitive for me. I expected `*.local.*` configs to be 
- of the highest priority of all config files, in order to override even ENV-specific configs files, for local (developer-specific) development.
- gitignored (again, because developer-specific)

(As far as I know that's how Sails does things, and I independently arrived at the same prioritization order in my own framework.)

As it stands however there is no way to, say force a DB port only in my personal development environment, if a ENV-specific config exists.
As I said this is pretty surprising. The only workaround that I can currently see is to set `NODE_ENV = 'local'`, which has it's own problems.

jbcpollak:
This issue is of intense debate in my team. We would like to use .local.* files for local-only configuration, however we need global / all .js configuration as well, which is not currently supported.

In addition, as @MarcDiethelm points, .local files should be the highest priority, and excluded from version control.

I would add that I would like to see .local files NOT stored in the server/ directory at all, Lumping config (especially version-control ignored config files) with source code seems like a recipe for confusion.

PR #125, opened 2015-04-20
Enable .js extension for master config files

phantom10111:
Some people (including myself) have run into the problem, that loopback-boot doesn't work properly with .js config files. After this modification, loopback-boot will try to load a .js file when a .json master config is not available. This includes model-config, which probably doesn't have much need for a .js file, but a possibility to have it is not a bad thing imo. The behaviour with .local and .{env} files is not changed (that is .js files are preferred to .json).

### DIFF
--- a/lib/config-loader.js
+++ b/lib/config-loader.js
@@ -31,7 +31,8 @@ ConfigLoader.loadDataSources = function(rootDir, env) {
  */
 ConfigLoader.loadModels = function(rootDir, env) {
   /*jshint unused:false */
-  return tryReadJsonConfig(rootDir, 'model-config') || {};
+  return tryReadJsonConfig(rootDir, 'model-config') ||
+         tryReadJsConfig(rootDir, 'model-config') || {};
 };
 
 /**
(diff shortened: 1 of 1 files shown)


slnode:

bajtos:
@slnode test please

bajtos:
@phantom10111 thank you for the pull request. I am afraid your patch will create a lot of headache in loopback-workspace, `slc loopback` and StrongLoop Arc, because they all expect a JSON file to be present. See #80 and #82 for more details.

Can you please rework the pull request to implement what I have proposed in #80?

bajtos:
@phantom10111 Ultimately we need to implement a different approach that will support `slc loopback` and `StrongLoop Arc`, as outlined in #80. However, as a short-term solution for your pain point, I am proposing to modify this pull request and add a flag (boot options) that will allow users to opt-in to the new behaviour.

An example usage of this new flag (feel free to come up with a better flag name):

```js
boot(app, {
  appRootDir: __dirname,
  masterConfigsInJavaScript: true  // false by default
});
```

Thoughts?

slnode:

bajtos:
@phantom10111 what's the status of this pull request? Are you still keen to get this feature landed?

slnode:

phantom10111:
I'm sorry, I currently don't have the time to handle the task properly. I may return to it at a later date but I'm closing the pull request for now.

PR #126, opened 2015-04-22
executor: fix port lookup

bajtos:
Use the global `isFinite` function to find the first valid port number. This is the same behaviour as `_.isFinite` in lodash@2.x.

As a result, only values that are a valid number are accepted, e.g. "0" is accepted but "0xy" is not.

This patch is fixing a regression introduced in 044a4df07a8c076ec0c72e98261fdb44631f7acd.

/to @raymondfeng please review


raymondfeng:
Lgtm

PR #128, opened 2015-04-23
Upgrade to lodash 3.x

rmg:
Replace single-function module with full lodash, which can be shared with other packages if installed in a parent module.

See strongloop/strongloop#213 for context. This is an interim solution so that loopback-workspace can benefit from this change without having to upgrade to loopback-boot 2.x.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -1,5 +1,5 @@
 var assert = require('assert');
-var cloneDeep = require('lodash.clonedeep');
+var cloneDeep = require('lodash/lang/cloneDeep');
 var fs = require('fs');
 var path = require('path');
 var ConfigLoader = require('./config-loader');
--- a/test/browser.test.js
+++ b/test/browser.test.js
@@ -8,6 +8,7 @@ var vm = require('vm');
 
 describe('browser support', function() {
   this.timeout(60000); // 60s to give browserify enough time to finish
+  beforeEach(sandbox.reset);
 
   it('has API for bundling and executing boot instructions', function(done) {
     var appDir = path.resolve(__dirname, './fixtures/browser-app');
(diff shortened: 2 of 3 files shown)


test/browser.test.js
```diff
@@ -8,6 +8,7 @@ var vm = require('vm');
 
 describe('browser support', function() {
   this.timeout(60000); // 60s to give browserify enough time to finish
+  sandbox.reset();
```
bajtos: This is wrong. It should be either `before(sandbox.reset)` or `beforeEach(sandbox.reset)`. I think beforeEach makes more sense.

lib/compiler.js
```diff
@@ -1,5 +1,5 @@
 var assert = require('assert');
-var cloneDeep = require('lodash.clonedeep');
+var cloneDeep = require('lodash/lang/cloneDeep');
```
bajtos: FWIW, compiler is not bundled for the browser, so the following is acceptable too: 
```js
var cloneDeep = require('lodash').cloneDeep;
```

bajtos:
Releasing this fix will be tricky, as npmjs.org displays README from the last published version. After we release 1.x, we must release another 2.x version to get the right README displayed.

test/browser.test.js
rmg: Yep, that's clearly wrong. I blame tap.

lib/compiler.js
rmg: Good point. I'll leave it as is though in case anyone reading gets curious about it and then accidentally learns something cool about lodash. ;-)

rmg:
@bajtos I fixed the `beforeEach()`. Ok to rebase and merge?

rmg:
Regarding a release, it looks like there are a couple unreleased changes on master.. if you do a 1.x release just before releasing them that would be awesome :-)

rmg:
Nevermind, my repo was out of date.. you released from master 8 hours ago.

bajtos:
LGTM

PR #130, opened 2015-04-29
Port can't be number checked to support iisnode

jsheely:
Using a parseInt() or number isNumber function won't work if we want to support iisnode which uses named pipes for ports (ex. \\.\pipe\mypipe)

### DIFF
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -406,10 +406,10 @@ describe('executor', function() {
       assert.equal(app.get('port'), 3000);
     });
 
-    it('should ignore non-numeric port values in ENV', function() {
-      process.env.PORT = '123invalid';
+    it('should respect named pipes port values in ENV', function() {
+      process.env.PORT = '\.\pipe\test';
       boot.execute(app, someInstructions({ config: { port: 3000 } }));
-      assert.equal(app.get('port'), 3000);
+      assert.equal(app.get('port'), '\.\pipe\test');
     });
   });
 
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -115,7 +115,9 @@ function setPort(app, instructions) {
     process.env.npm_package_config_port,
     app.get('port'),
     3000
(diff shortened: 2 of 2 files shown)


slnode:

bajtos:
Good catch, thank you for the contribution. Is there any reasonably easy way how to write automated test for this?

Do you happen to know if named pipes work on Unix platforms too? (Perhaps @bnoordhuis will know?)

bajtos:
@slnode ok to test

bajtos:
@jsheely please make the CI build pass, it fails on this test now:


```
executor with PaaS and npm env variables should ignore non-numeric port values in ENV:
     AssertionError: "123invalid" == 3000
```

jsheely:
@bajtos Evidently it is a unix and windows thing. However as you can read they work a little differently.

http://en.wikipedia.org/wiki/Named_pipe

I'll take a look at the build CI

bnoordhuis:
Sorry @bajtos, missed your mention.

@jsheely already pointed out the similarities.  I'll just add that FIFOs are semantically a bit closer to Windows named pipes but the closest thing that is easy to work with in io.js and node.js, is a UNIX domain socket.

test/executor.test.js
```diff
@@ -406,10 +406,10 @@ describe('executor', function() {
       assert.equal(app.get('port'), 3000);
     });
 
-    it('should ignore non-numeric port values in ENV', function() {
-      process.env.PORT = '123invalid';
+    it('should respect named pipes port values in ENV', function() {
+      process.env.PORT = '\.\pipe\test';
```
bajtos: I believe you need to encode backslashes as `\\.pipe\\test`.

bajtos:
I have fixed the pipe name string in the test, landed the patch via 083a126 and released as `loopback-boot@v2.8.0`.

Thank you for the contribution!

PR #131, opened 2015-04-30
support 'mixinsources' option

PradnyaBaviskar:
In continuation with - https://github.com/strongloop/loopback-boot/issues/79, added support for ```mixinSources``` option.

The implementation -
- By default looks for mixinsources in ```mixins``` directory
- Loads only mixins used through models

Connect to #79

@bajtos - Can you review?

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -85,8 +85,9 @@ module.exports = function compile(options) {
     modelsRootDir, modelsConfig, modelSources);
 
   var mixinDirs = options.mixinDirs || [];
+  var mixinSources = options.mixinSources || modelsMeta.mixins || ['./mixins'];
   var mixinInstructions = buildAllMixinInstructions(
-    appRootDir, mixinDirs, options);
+    appRootDir, mixinDirs, mixinSources, options, modelInstructions);
 
   // When executor passes the instruction to loopback methods,
   // loopback modifies the data. Since we are loading the data using `require`,
@@ -606,12 +607,48 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
   return (fixedFile === undefined ? resolvedPath : fixedFile);
 }
 
-function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
+function buildAllMixinInstructions(appRootDir, mixinDirs, mixinSources, options,
+  modelInstructions) {
   var extensions = _.without(_.keys(require.extensions),
     _.keys(getExcludedExtensions()));
-  var files = options.mixins || [];
 
-  mixinDirs.forEach(function(dir) {
+  // load mixins from `options.mixins`
+  var sourceFiles = options.mixins || [];
+  var instructionsFromMixins = loadMixins(sourceFiles, options);
+
+  // load mixins from `options.mixinDirs`
+  sourceFiles = findMixinDefinitions(appRootDir, mixinDirs, extensions);
+  if (sourceFiles === undefined) return;
+  var instructionsFromMixinDirs = loadMixins(sourceFiles, options);
+
+  /* If `mixinDirs` and `mixinSources` have any directories in common,
+   * then remove the common directories from `mixinSources` */
+  mixinSources = _.difference(mixinSources, mixinDirs);
+
+  // load mixins from `options.mixinSources`
+  sourceFiles = findMixinDefinitions(appRootDir, mixinSources, extensions);
+  if (sourceFiles === undefined) return;
+  var instructionsFromMixinSources = loadMixins(sourceFiles, options);
+
+  // Fetch unique list of mixin names, used in models
+  var modelMixins = fetchMixinNamesUsedInModelInstructions(modelInstructions);
+  modelMixins = _.unique(modelMixins);
+
+  // Filter-in only mixins, that are used in models
+  instructionsFromMixinSources = filterMixinInstructionsUsingWhitelist(
+    instructionsFromMixinSources, modelMixins);
+
+  var mixins = _.assign(
+    instructionsFromMixins,
+    instructionsFromMixinDirs,
+    instructionsFromMixinSources);
+
+  return _.values(mixins);
+}
+
+function findMixinDefinitions(appRootDir, sourceDirs, extensions) {
+  var files = [];
+  sourceDirs.forEach(function(dir) {
     dir = tryResolveAppPath(appRootDir, dir);
     if (!dir) {
       debug('Skipping unknown module source dir %j', dir);
@@ -619,8 +656,12 @@ function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
     }
     files = files.concat(findScripts(dir, extensions));
   });
+  return files;
+}
 
-  var mixins = files.map(function(filepath) {
+function loadMixins(sourceFiles, options) {
+  var mixinInstructions = {};
+  sourceFiles.forEach(function(filepath) {
     var dir = path.dirname(filepath);
     var ext = path.extname(filepath);
     var name = path.basename(filepath, ext);
@@ -634,10 +675,31 @@ function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
       _.extend(meta, require(metafile));
     }
     meta.sourceFile = filepath;
-    return meta;
+    mixinInstructions[meta.name] = meta;
   });
 
-  return mixins;
+  return mixinInstructions;
+}
+
+function fetchMixinNamesUsedInModelInstructions(modelInstructions) {
+  return _.flatten(modelInstructions
+  .map(function(model) {
+    return model.definition && model.definition.mixins ?
+      Object.keys(model.definition.mixins) : [];
+  }));
+}
+
+function filterMixinInstructionsUsingWhitelist(instructions, includeMixins) {
+  var instructionKeys = Object.keys(instructions);
+  includeMixins = _.intersection(instructionKeys, includeMixins);
+
+  var filteredInstructions = {};
+  instructionKeys.forEach(function(mixinName) {
+    if (includeMixins.indexOf(mixinName) !== -1) {
+      filteredInstructions[mixinName] = instructions[mixinName];
+    }
+  });
+  return filteredInstructions;
 }
 
 function normalizeMixinName(str, options) {
--- a/test/compiler.test.js
+++ b/test/compiler.test.js
@@ -1016,41 +1016,175 @@ describe('compiler', function() {
     });
 
     describe('for mixins', function() {
-      function verifyMixinIsFoundViaMixinDirs(sourceFile, mixinDirs) {
-        var appJS = appdir.writeFileSync(sourceFile, '');
+      describe(' - mixinDirs', function() {
+        function verifyMixinIsFoundViaMixinDirs(sourceFile, mixinDirs) {
+          var appJS = appdir.writeFileSync(sourceFile, '');
 
-        var instructions = boot.compile({
-          appRootDir: appdir.PATH,
-          mixinDirs: mixinDirs
+          var instructions = boot.compile({
+            appRootDir: appdir.PATH,
+            mixinDirs: mixinDirs
+          });
+
+          expect(instructions.mixins[0].sourceFile).to.eql(appJS);
+        }
+
+        it('supports `mixinDirs` option', function() {
+          verifyMixinIsFoundViaMixinDirs('custom-mixins/other.js',
+            ['./custom-mixins']);
         });
 
-        expect(instructions.mixins[0].sourceFile).to.eql(appJS);
-      }
+        it('resolves relative path in `mixinDirs` option', function() {
+          verifyMixinIsFoundViaMixinDirs('custom-mixins/other.js',
+            ['./custom-mixins']);
+        });
 
-      it('supports `mixinDirs` option', function() {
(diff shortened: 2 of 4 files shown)


lib/compiler.js
```diff
@@ -637,9 +659,25 @@ function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
     return meta;
   });
 
+  if (includeMixins.length > 0) {
+    mixins = mixins.filter(function(m) {
+      return (includeMixins.indexOf(m.name) === -1 ? false : true);
+    });
+  }
```
bajtos: I am proposing to move this filtering code out of "loadMixins" to keep "loadMixins" simpler and focused on doing one thing only.

lib/compiler.js
```diff
@@ -637,9 +659,25 @@ function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
...
 }
 
+function fetchMixinsUsedInModels(modelInstructions) {
+  var modelMixins = [];
+  modelInstructions.forEach(function(model) {
+    if (model.definition && model.definition.mixins) {
+      modelMixins = modelMixins.concat(Object.keys(model.definition.mixins));
+    }
+  });
+  return modelMixins;
```
bajtos: You can use [_.flatten](https://lodash.com/docs#flatten) to make the filtering code simpler.

```js
return _(modelInstructions)
  .map(function(model) {
    return model.definition && model.definition.mixins ? Object.keys(model.definition.mixins) : [];
  })
  .flatten();
```

test/compiler.test.js
```diff
@@ -1016,41 +1016,106 @@ describe('compiler', function() {
...
-      it('supports `mixinDirs` option', function() {
-        verifyMixinIsFoundViaMixinDirs('mixins/other.js', ['./mixins']);
+        it('resolves module relative path in `mixinDirs` option', function() {
+          verifyMixinIsFoundViaMixinDirs('node_modules/custom-mixins/other.js',
+            ['custom-mixins']);
+        });
       });
 
-      it('resolves relative path in `mixinDirs` option', function() {
-        verifyMixinIsFoundViaMixinDirs('custom-mixins/vehicle.js',
```
bajtos: Is there any particular reason why `vehicle.js` was renamed to `other.js`?

test/compiler.test.js
```diff
@@ -1016,41 +1016,106 @@ describe('compiler', function() {
...
+      describe(' - mixinSources', function() {
+        function verifyMixinIsFoundViaMixinSources(sourceFile, mixinSources) {
+          var appJS = appdir.writeFileSync(sourceFile, '');
+
+          var instructions = boot.compile({
+            appRootDir: appdir.PATH,
+            mixinSources: mixinSources
+          });
+
+          expect(instructions.mixins[0].sourceFile).to.eql(appJS);
```
bajtos: If my understanding is correct, then this function should always fail if the compiler implementation works correctly. The mixin instruction should be included only when there is a model using the mixin. Since there are no models defined here, the compiler should not load any mixins from `mixinSources`.

Am I missing something?

test/compiler.test.js
PradnyaBaviskar: It was not fitting the 80 character limit. So, instead of moving ```custom-mixins/vehicle.js``` to new line, I renamed ```vehicle.js``` to ```other.js```

bajtos: Haha, fair enough :)

test/compiler.test.js
PradnyaBaviskar: Yes, you are right!
My code to filter mixinSources is currently 
```
if (includeMixins.length > 0) {
   ...
}
```
The condition is not needed. I'll fix this!

lib/compiler.js
```diff
@@ -606,16 +607,37 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
...
+
+  files = [];
+  var mixinsFromMixinSources = loadMixins(appRootDir, mixinSources, files,
+    extensions, options);
+  if (mixinsFromMixinSources === undefined) return;
+
+  // Filter in only mixins, that are used in models
+  mixinsFromMixinSources = filterMixins(mixinsFromMixinSources, modelMixins);
+
+  mixins = mixins.concat(mixinsFromMixinSources);
```
bajtos: This may create duplicate items in the `mixins` array, when the same mixin is loaded from `mixinDirs` and `mixinSources`, is that correct? In which case you should de-duplicate the instructions.

lib/compiler.js
```diff
@@ -640,6 +662,21 @@ function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
...
 
+function fetchMixinsUsedInModels(modelInstructions) {
+  return _.flatten(modelInstructions
+  .map(function(model) {
+    return model.definition && model.definition.mixins ?
+      Object.keys(model.definition.mixins) : [];
+  }));
+}
+
+function filterMixins(mixinInstructions, includeMixins) {
```
bajtos: Please use a more descriptive function name, e.g. `filterMixinInstructionsUsingWhitelist`.

lib/compiler.js
```diff
@@ -640,6 +662,21 @@ function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
   return mixins;
 }
 
+function fetchMixinsUsedInModels(modelInstructions) {
```
bajtos: Please use different names for "a list of mixin instructions" and "a list of mixin names". Since there are no types in javascript, we must be more careful about method naming to make their purpose clear.

In this case, I am proposing `fetchMixinNamesUsedInModelInstructions`.

test/executor.test.js
```diff
@@ -306,15 +307,25 @@ describe('executor', function() {
...
-        var options = {
-          appRootDir: appdir.PATH,
-          mixinDirs: ['./mixins']
+        options = {
+          appRootDir: appdir.PATH
         };
+      });
 
+      it('defines mixins from instructions - using `mixinDirs`', function() {
+        options.mixinDirs = ['./mixins'];
```
bajtos: Hmm, once `mixinSources` are correctly implemented, this test always passes, because `./mixins` is the default value of `mixinSources`. Isn't it so? I think you need to change this test to use `./custom-mixins` instead.

PradnyaBaviskar:
@bajtos - Addressed your comments above and have also added a few more test cases.

Can you please review?

lib/compiler.js
```diff
@@ -606,21 +607,51 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
...
 
-function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
+function buildAllMixinInstructions(appRootDir, mixinDirs, mixinSources, options,
+  modelInstructions) {
   var extensions = _.without(_.keys(require.extensions),
     _.keys(getExcludedExtensions()));
+
   var files = options.mixins || [];
+  var mixins = loadMixins(appRootDir, mixinDirs, files, extensions, options);
+  if (mixins === undefined) return;
```
bajtos: I find this difficult to grasp at the first sight, it looks like `loadMixins` should be modifying the `files` variable.

lib/compiler.js
```diff
@@ -606,21 +607,51 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
...
+  mixins = _.reject(mixins, function(instruction) {
+    return (_.findWhere(mixinsFromMixinSources,
+      {name: instruction.name}) !== undefined);
+  });
+  mixins = mixins.concat(mixinsFromMixinSources);
+  return mixins;
+}
 
-  mixinDirs.forEach(function(dir) {
+function loadMixins(appRootDir, sourceDirs, files, extensions, options) {
```
bajtos: Perhaps rename `files` to something like `sourceFiles` or `extraSourceFiles` and just pass `options.mixins` when calling this method?

lib/compiler.js
```diff
@@ -606,21 +607,51 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
...
   var files = options.mixins || [];
+  var mixins = loadMixins(appRootDir, mixinDirs, files, extensions, options);
+  if (mixins === undefined) return;
+
+  /* If `mixinDirs` and `mixinSources` have any directories in common,
+   * then remove the common directories from `mixinSources` */
+  mixinSources = _.difference(mixinSources, mixinDirs);
+  files = [];
+  var mixinsFromMixinSources = loadMixins(appRootDir, mixinSources, files,
+    extensions, options);
```
bajtos: Actually this is the place I find confusing regarding the `files` variable. If we go with the proposal in my previous comment, then this should become `loadMixins(appRootDir, mixinSources, null, extensions, options)`.

Or even better, drop the `files` (`sourceFiles`) parameter completely and implement the concatenation at the instructions level.

lib/compiler.js
```diff
@@ -606,21 +607,51 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
...
+}
 
-  mixinDirs.forEach(function(dir) {
+function loadMixins(appRootDir, sourceDirs, files, extensions, options) {
+  sourceDirs.forEach(function(dir) {
     dir = tryResolveAppPath(appRootDir, dir);
     if (!dir) {
       debug('Skipping unknown module source dir %j', dir);
-      return;
+      return undefined;
```
bajtos: The return value is ignored by `forEach`, isn't it?

lib/compiler.js
```diff
@@ -634,10 +665,26 @@ function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
...
+function fetchMixinNamesUsedInModelInstructions(modelInstructions) {
+  return _.flatten(modelInstructions
+  .map(function(model) {
+    return model.definition && model.definition.mixins ?
+      Object.keys(model.definition.mixins) : [];
+  }));
+}
+
+function filterMixinInstructionsUsingWhitelist(mixinInstructions,
+  includeMixins) {
```
bajtos: Perhaps `(instructions, names)` would fit into 80 chars? If not, then please indent the second line by two stops to provide at least some sort of visual distinction between the arg list and the function body.

```js
function(....
    ...) {
  // code
}
```

bajtos:
The implementation looks good in general now, although I find it difficult to understand the implementation in compiler. Can you please try to simplify it a bit more, using my comments above for inspiration? Don't worry about it too much. If the result does not look better than the current solution or it takes too much time, then I am ok to land the patch as it is now.

PradnyaBaviskar:
@bajtos - Done with the changes suggested by you. Please review.

lib/compiler.js
```diff
@@ -606,21 +607,59 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
...
+  instructionsFromMixinSources = filterMixinInstructionsUsingWhitelist(
+    instructionsFromMixinSources, modelMixins);
+
+  var mixins = concatMixinInstructions(instructionsFromMixins,
+    instructionsFromMixinDirs);
+  mixins = concatMixinInstructions(mixins, instructionsFromMixinSources);
+  return mixins;
+}
+
+function findMixinDefinition(appRootDir, sourceDirs, extensions) {
```
bajtos: nitpick: please rename to `findMixinDefinitions` (plural).

lib/compiler.js
bajtos: Please revert this line, the return value is ignored by `forEach` and `return` is the same as `return undefined`.

lib/compiler.js
```diff
@@ -634,9 +673,33 @@ function buildAllMixinInstructions(appRootDir, mixinDirs, options) {
...
+      return (includeMixins.indexOf(m.name) === -1 ? false : true);
+    });
+}
+
+function concatMixinInstructions(rejectionSet, inclusionSet) {
+  var mixins = _.reject(rejectionSet, function(instruction) {
+    return (_.findWhere(inclusionSet,
+      {name: instruction.name}) !== undefined);
+  });
+  mixins = mixins.concat(inclusionSet);
```
bajtos: This function is pretty complicated and difficult to understand. I am proposing the following change:

 - modify `loadMixins` to return a hashmap (lookup table), not an array
 - replace `concatMixinInstructions` with

   ```js
_.assign(instructionsFromMixinSources, instructionsFromMixinDirs, instructionsFromMixins)
```

Thoughts?

PradnyaBaviskar:
Reworked on ```loadMixins```, so that it returns lookup table, and not an array.

bajtos:
Awesome, landed!

ramshgithub:
I am facing an issue as "column "columnname" does not exist", when i try to find any data from my Person table.I have 'name' and 'role' fields in my Person table, and 'status' is not present in my table(not require same in table).But my model.json have name,role and status as i needed all these fields in my explorer API.how can i solve this error?this happens whenever we have extra column name in model which is not present in our DB

Person.json
"name": "Person",
"base": "PersistedModel",
"idInjection": true,
"options": {
"validateUpsert": true
},
"properties": {
"name": {
"type": "string"
},
"status": {
"type": "string"
},
"role": {
"type": "string"
}
},
"validations": [],
"relations": {
"products": {
"type": "hasMany",
"model": "Product",
"foreignKey": "personId",
"description": "Returns a list of products associated with a person."
}
},
"methods": {}
}

bajtos:
@ramshgithub AFAICT, your issue is completely unrelated to 'mixinsources' option implemented by this pull request. Please open a new issue in LoopBack's main issue tracker: https://github.com/strongloop/loopback/issues/new

Issue #75, opened 2014-11-28
Declaratively load loopback.rest and loopback-explorer


bajtos:
In order to remove `boot/rest` and `boot/explorer` from the api-server template, we need to extend loopback-boot to support the following features:

 1. Specifying middleware parameters as a value to be read via `app.get` from `config.json`, example:

  ```json
  {
    "routes": {
      "loopback#rest": {
        "paths": ["${restApiRoot}"]
      }
  }
  ```

 2. Allowing the middleware to be optional, e.g.

  ```json
  {
    "routes": {
      "loopback-explorer": {
        "optional": "Run `npm install loopback-explorer` to enable the LoopBack explorer.",
        "params": {
          "basePath": "$!{restApiRoot}"
        }
     }
  }
  ```     

 3. Rework the explorer so that it can get the `app` object via `req.app` on the first request, in the same way as `loopback.rest`.

hacksparrow:
Let's address the issue "one feature at a time". I think **2. Allowing the middleware to be optional** is the most straightforward one, so let's give it the first shot.

So @bajtos what are the requirements for **2**? Here is what I am assuming:

1. If the middleware is not resolvable, don't `throw` and resume loading the rest of the middleware. Also, print the `optional` message to stdio.
2. If the middleware is resolvable, load it like other middleware.

Please confirm or correct.

bajtos:
> 1. If the middleware is not resolvable, don't `throw` and resume loading the rest of the middleware. Also, print the `optional` message to stdio.
> 2. If the middleware is resolvable, load it like other middleware.

@hacksparrow that's perfect :+1: 

> Let's address the issue "one feature at a time".

+1 I'd add "one pull request per feature", but feel free to use whatever workflow that works best for you.

hacksparrow:
@bajtos I have implemented a way to to pass `app` as a parameter to middleware via `middleware.json`. `app` can be referenced as `${app}`.

**middleware.json**
```
"routes": {
  "./middleware/hello": {
    "params": ["${app}", {"app": "${app}"}]
  }
}
```
How does the convention `${app}` look like to you?

hacksparrow:
@bajtos also, for explorer to work seamlessly as a self-contained optional middleware, the middleware config should be the following, if I am not wrong:

```
{
  "routes": {
    "loopback-explorer": {
      "optional": "Run `npm install loopback-explorer` to enable the LoopBack explorer.",
      "params":  ["${app}", {
[...]

PR #134, opened 2015-06-08
allow middleware to be optional

hacksparrow:
Added two test cases and implemented the feature as discussed in #75. The implementation made both the tests pass.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -469,8 +469,12 @@ function buildMiddlewareInstructions(rootDir, config) {
 
         var resolved = resolveMiddlewarePath(rootDir, middleware, config);
 
-        // resolved.sourceFile will be `false` if an optional middleware is not resolvable
-        // if a non-optional middleware is not resolvable, it will throw at resolveAppPath()
+        /*
+        * resolved.sourceFile will be `false` if an optional middleware
+        * is not resolvable.
+        * if a non-optional middleware is not resolvable, it will throw
+        * at resolveAppPath() and not reach here
+        */
         if (resolved.sourceFile) {
           var middlewareConfig = cloneDeep(config);
           middlewareConfig.phase = phase;
@@ -487,9 +491,8 @@ function buildMiddlewareInstructions(rootDir, config) {
           if (resolved.fragment) {
             item.fragment = resolved.fragment;
           }
-          middlewareList.push(item);          
-        }
-        else {
+          middlewareList.push(item);
+        } else {
           debug('Middleware "%s" not found: %s', middleware, resolved.optional);
         }
 
@@ -665,8 +668,7 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
     var files = tryReadDir(sourceDir);
     var fixedFile = fixFileExtension(resolvedPath, files, false);
     return (fixedFile === undefined ? resolvedPath : fixedFile);
-  }
-  else {
+  } else {
     return false;
   }
 }
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -451,22 +451,16 @@ function loadModelDefinition(rootDir, jsonFile, allFiles) {
 }
 
 function buildMiddlewareInstructions(rootDir, config) {
-
   var phasesNames = Object.keys(config);
   var middlewareList = [];
-
   phasesNames.forEach(function(phase) {
-
     var phaseConfig = config[phase];
-
     Object.keys(phaseConfig).forEach(function(middleware) {
-
       var allConfigs = phaseConfig[middleware];
       if (!Array.isArray(allConfigs))
         allConfigs = [allConfigs];
 
       allConfigs.forEach(function(config) {
-
         var resolved = resolveMiddlewarePath(rootDir, middleware, config);
 
         /*
@@ -498,9 +492,7 @@ function buildMiddlewareInstructions(rootDir, config) {
           item.fragment = resolved.fragment;
         }
         middlewareList.push(item);
-
       });
-
     });
   });
 
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -352,7 +352,7 @@ function resolveAppPath(rootDir, relativePath, resolveOptions) {
   if (resolvedPath === undefined) {
 
     if (resolveOptions.optional) {
-      return false
+      return false;
     }
 
     var err = new Error('Cannot resolve path "' + relativePath + '"');
@@ -490,7 +490,7 @@ function buildMiddlewareInstructions(rootDir, config) {
           middlewareList.push(item);          
         }
         else {
-          debug('Middleware "%s" not found: %s', middleware, resolved.optional)
+          debug('Middleware "%s" not found: %s', middleware, resolved.optional);
         }
 
       });
@@ -537,7 +537,7 @@ function resolveMiddlewarePath(rootDir, middleware, config) {
 
   if (!fragment) {
     resolved.sourceFile = resolveAppScriptPath(rootDir, middlewarePath, opts);
-    resolved.optional = config.optional
+    resolved.optional = config.optional;
     return resolved;
   }
 
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -349,12 +349,7 @@ function findModelDefinitions(rootDir, sources) {
 
 function resolveAppPath(rootDir, relativePath, resolveOptions) {
   var resolvedPath = tryResolveAppPath(rootDir, relativePath, resolveOptions);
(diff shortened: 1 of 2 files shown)


hacksparrow:
Yo @bajtos! Please take a look.

Jenkins is failing again without giving any reasons; I have encountered this on another repo earlier. Any idea what's going on?

lib/compiler.js
```diff
@@ -350,6 +350,11 @@ function findModelDefinitions(rootDir, sources) {
 function resolveAppPath(rootDir, relativePath, resolveOptions) {
   var resolvedPath = tryResolveAppPath(rootDir, relativePath, resolveOptions);
   if (resolvedPath === undefined) {
+
+    if (resolveOptions.optional) {
+      return false;
+    }
+
```
bajtos: I'd personally simplify L352-L357 to the following:

```
if (resolvedPath === undefined && !resolveOptions.optional) {
  // report error
}
return resolvedPath; // returns undefined, which is false-y too
```

lib/compiler.js
```diff
@@ -461,25 +466,38 @@ function buildMiddlewareInstructions(rootDir, config) {
...
+          var item = {
+            sourceFile: resolved.sourceFile,
+            config: middlewareConfig
+          };
+          if (resolved.fragment) {
+            item.fragment = resolved.fragment;
+          }
+          middlewareList.push(item);
+        } else {
+          debug('Middleware "%s" not found: %s', middleware, resolved.optional);
```
bajtos: Please use a reverse condition and an early return, that will reduce the amount of indentation changes in the diff. Also use `console.log` instead of `debug` to print the message.

```js
if (!resolved.sourceFile) {
  // print message;
  return; // from forEach()
}

// proceed with the original code
```

lib/compiler.js
hacksparrow: Makes sense.

lib/compiler.js
hacksparrow: Got it.

lib/compiler.js
```diff
@@ -499,14 +517,17 @@ function buildMiddlewareInstructions(rootDir, config) {
...
   var resolved = {};
 
   var segments = middleware.split('#');
   var pathName = segments[0];
   var fragment = segments[1];
   var middlewarePath = pathName;
-  var opts = { strict: true };
+  var opts = {
+    strict: true,
+    optional: config.optional
```
bajtos: `config.optional` is a string, `opts.optional` is boolean. Perhaps it's better to coerce the value here?

```js
optional: !!config.optional
```

lib/compiler.js
```diff
@@ -641,10 +663,14 @@ function fixFileExtension(filepath, files, onlyScriptsExportingFunction) {
...
-  var fixedFile = fixFileExtension(resolvedPath, files, false);
-  return (fixedFile === undefined ? resolvedPath : fixedFile);
+  if (resolvedPath) {
+    var sourceDir = path.dirname(resolvedPath);
+    var files = tryReadDir(sourceDir);
+    var fixedFile = fixFileExtension(resolvedPath, files, false);
+    return (fixedFile === undefined ? resolvedPath : fixedFile);
+  } else {
+    return false;
+  }
```
bajtos: Please use a reverse condition and an early return:

```js
var resolvedPath = resolveAppPath(...);
if (!resolvedPath) {
  return resolvedPath;
}
// original code stays unchanged
```

test/compiler.test.js
```diff
@@ -1564,6 +1564,21 @@ describe('compiler', function() {
         .to.throw(/path-does-not-exist/);
     });
 
+    it(
+      'does not fail when an optional middleware cannot be resolved',
+      function() {
```
bajtos: Please keep the formatting consistent with the code around, see e.g. the test below.

```js
it('does not fail...',
  function() {
    // ...
  });
```

test/compiler.test.js
```diff
@@ -1578,6 +1593,23 @@ describe('compiler', function() {
           .to.throw(/path-does-not-exist/);
       });
 
+    it(
+      'does not fail when an optional middleware fragment cannot be resolved',
+      function() {
```
bajtos: ditto

lib/compiler.js
hacksparrow: I used the value assuming, it might be needed somewhere down the line. Value coercion should suffice in the current implementation. Will change it.

test/compiler.test.js
hacksparrow: jscs was complaining about the 80 characters things. Is the 80 characters limit an intentional choice?

bajtos:
Hi @hacksparrow, thank you for the pull request!

The Jenkins CI build is complaining about linter errors, which is rather weird, since you have already fixed them in additional commits. I'll restart the build.

bajtos:
@slnode test please

test/compiler.test.js
bajtos: > jscs was complaining about the 80 characters things.

I see. Is the line still longer than 80 characters when you keep `it` + test name on the same line? I think it should fit 80 chars as long as `function() {` is on the next line.

>  Is the 80 characters limit an intentional choice?

Yes. First of all, the diff viewport on GitHub does not handle long lines well. One has to scroll horizontally, which is PITA. However, I think the viewport width is around 100 characters, so that does not justify the value 80.

I took 80 chars from Node's coding style. The limit is mostly historical, where it the goal was to support text terminals 80 chars x 25 lines, but in practice it works well on modern computers too. For example it allows me to edit two files side-by-side.

hacksparrow: Using just "middleware" instead of "module middleware" does not cross the 80 characters limit. Will update the formatting accordingly.

And thanks for the info on the 80 characters limit.

lib/compiler.js
```diff
@@ -451,17 +451,36 @@ function loadModelDefinition(rootDir, jsonFile, allFiles) {
 }
 
 function buildMiddlewareInstructions(rootDir, config) {
+
```
bajtos: extra empty line, please remove

lib/compiler.js
```diff
@@ -451,17 +451,36 @@ function loadModelDefinition(rootDir, jsonFile, allFiles) {
...
   var phasesNames = Object.keys(config);
   var middlewareList = [];
+
   phasesNames.forEach(function(phase) {
+
```
bajtos: extra empty line, please remove

lib/compiler.js
```diff
@@ -451,17 +451,36 @@ function loadModelDefinition(rootDir, jsonFile, allFiles) {
...
     var phaseConfig = config[phase];
+
     Object.keys(phaseConfig).forEach(function(middleware) {
+
```
bajtos: extra empty line, please remove

lib/compiler.js
```diff
@@ -451,17 +451,36 @@ function loadModelDefinition(rootDir, jsonFile, allFiles) {
...
       var allConfigs = phaseConfig[middleware];
       if (!Array.isArray(allConfigs))
         allConfigs = [allConfigs];
 
       allConfigs.forEach(function(config) {
-        var resolved = resolveMiddlewarePath(rootDir, middleware);
+
```
bajtos: Extra empty line, please remove.

In general, we try to avoid whitespace-only changes as they clutter the git history.

lib/compiler.js
```diff
@@ -479,7 +498,9 @@ function buildMiddlewareInstructions(rootDir, config) {
           item.fragment = resolved.fragment;
         }
         middlewareList.push(item);
+
```
bajtos: extra empty line, please remove

lib/compiler.js
```diff
@@ -479,7 +498,9 @@ function buildMiddlewareInstructions(rootDir, config) {
...
       });
+
```
bajtos: extra empty line, please remove

crandmck:
@bajtos Could you please review this PR and merge if it looks OK?  Hage is waiting on this to move ahead with some other work...Thanks!

bajtos:
Sure. 

BTW GitHub does not send any notifications when new commits are added, one has to add a comment to notify reviewers.

bajtos:
@hacksparrow the commit 272fb33 is very suspicious, how did you resolve the conflict? The correct way is to run `git rebase master && git push -f`. Don't worry about it now, I'll clean up the history myself.

hacksparrow:
@bajtos the conflict was in just a line, looks like I missed something. Thanks for the tip!

bajtos:
I did few more cleanups and landed the patch via ac5265f. Thank you for the contribution!

PR #136, opened 2015-06-24
Excl. mod. main path from middleware instructions

bajtos:
When loading middleware configured in "middleware.json", loopback-boot
used to require() the full path to middleware's module main file,
for example "node_modules/strong-express-metrics/index.js".

This is breaking strong-agent probes, as it's difficult to detect
when the intercepted module (e.g. "strong-express-metrics") was loaded.

This commit modifies the compiler to omit the tail (the path to the
module main source file) when requiring middleware.

Connect to strongloop-internal/scrum-loopback#279
See also https://github.com/strongloop/strongops/issues/278 which is fixing strong-agent.

/to @raymondfeng @sam-github please review

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -408,21 +408,29 @@ function tryResolveAppPath(rootDir, relativePath, resolveOptions) {
 
   fullPath = modulePaths
     .map(function(candidateDir) {
+      var absPath = path.join(candidateDir, relativePath);
       try {
-        var filePath = path.join(candidateDir, relativePath);
-        filePath = require.resolve(filePath);
+        // NOTE(bajtos) We need to create a proper String object here,
+        // otherwise we can't attach additional properties to it
+        /*jshint -W053 */
+        var filePath = new String(require.resolve(absPath));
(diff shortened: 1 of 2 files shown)


raymondfeng:
Is it an issue if the `package.json` has a `main` set to `lib/server.js`?

bajtos:
> Is it an issue if the package.json has a main set to lib/server.js?

It depends on what you mean by "is it an issue". This patch does not care what is the value of "package.main". When I oversimplify, we stop just before looking at "package.main".

Here is how the algorithm works:
 1. gather all "node_modules" folders that would be scanned by node
 2. for each folder, build a path "folder/middleware-name"
 3. for each path built in 2, try to require.resolve("folder/middleware-name")
 4. return the first path that actually exists

With this patch in place, the step will return the value built in step 2, not the full path built in step 3.

Is it more clear now?

raymondfeng:
Thank you for the explanation. LGTM.

bajtos:
Released as `loopback-boot@2.8.2`.

PR #140, opened 2015-07-08
Enable model-config.json _meta.mixinDirs

fabien:
Hi @bajtos - it's been I while since I've booted from a base loopback install (I usually started my own boot and passed in the options there). Today I've come to realize that there is currently no way to specify `mixinDirs` from `model-config.json`. 

I'm not that happy with `_meta.mixins` vs. `_meta.mixinDirs`, but it looks like the only option right now. Ideally, I would like to just have it point to `common/mixins` by default, but that's probably a no-go, because such entries are not in loopback-boot, but in said `model-config.json`.

What are your thoughts?

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -84,7 +84,7 @@ module.exports = function compile(options) {
   var modelInstructions = buildAllModelInstructions(
     modelsRootDir, modelsConfig, modelSources, options.modelDefinitions);
 
-  var mixinDirs = options.mixinDirs || [];
+  var mixinDirs = options.mixinDirs || modelsMeta.mixinDirs || [];
   var mixinSources = options.mixinSources || modelsMeta.mixins || ['./mixins'];
   var mixinInstructions = buildAllMixinInstructions(
     appRootDir, mixinDirs, mixinSources, options, modelInstructions);


bajtos:
> I'm not that happy with _meta.mixins vs. _meta.mixinDirs, but it looks like the only option right now. Ideally, I would like to just have it point to common/mixins by default, but that's probably a no-go, because such entries are not in loopback-boot, but in said model-config.json.

I am wondering why do you prefer to use `mixinDirs` instead of `mixinSources`? If all your models specify their mixins in their model JSON files, then I think `mixinSources` should work equally well for you.

I am not very happy about exposing `mixinDirs` via `model-config.json`, as it puts users in front of a difficult decision - which option to use, when it's better to use one option and when the other is better. So far, there was no decision to make.

Having said that, if we cannot find a better way how to solve your use-case, then I am ok with adding landing this patch, as long as there is a unit-test to verify the new feature. See https://github.com/strongloop/loopback-boot/blob/4f6e76842629c2fa7d897101c057c5622600345d/test/compiler.test.js#L1287-L1300 for inspiration.

bajtos:
Last but not least, could you please sign (again?) our CLA for this GH project? I am not sure why it did not recognize your handle.

bajtos:
@fabien ping, are you still keen to get this landed?

bajtos:
I am closing this pull request since it seems to be abandoned for very long time.

PR #142, opened 2015-07-29
config: support load via .js

yorkie:
as title

### DIFF
--- a/lib/config-loader.js
+++ b/lib/config-loader.js
@@ -87,7 +87,7 @@ function loadNamed(rootDir, env, name, mergeFn) {
  * @returns {Array.<String>} Array of absolute file paths.
  */
 function findConfigFiles(appRootDir, env, name) {
-  var master = ifExists(name + '.json');
+  var master = ifExists(name + '.json') || ifExists(name + '.js');
   if (!master) return [];
 
   var candidates = [


slnode:

raymondfeng:
Which json files do you want to allow js? A lot of them already supports it, for example, server/datasources.local.js for server/datasources.json.

yorkie:
Hey Feng, my use case is I wanna config some of fields in datasources in env-specific way, so actually I just want the default config looks like:

```js
module.exports = {
  mongo: {
    url: process.env.MONGO_URI
  }
};
```

The loopback config doesn't support the JavaScript(.js) file for the default config, this is what I need.

yorkie:
Without this, I need write duplicated and redundant config for this mongoDB, I just think that is great if loopback can load `server/datasources.js`.

raymondfeng:
You should be able to do that with `server/datasources.local.js` today.

yorkie:
If I specify `NODE_ENV=prod node server/app.js`, local is able to be loaded as well? I can't see something about local from your search algorithm actually.

raymondfeng:
datasources.local.js is loaded under all envs.

raymondfeng:
See http://docs.strongloop.com/display/LB/datasources.json#datasources.json-Environment-specificconfiguration

yorkie:
Thanks I see, I will close this :)

PR #144, opened 2015-08-03
Ensure all models are defined and attached before running custom scripts

ritch:
Make it easier to reference models that are loaded into the registry...

/to @raymondfeng 
/cc @bajtos

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -30,6 +30,7 @@ module.exports = function execute(app, instructions, callback) {
 
   setupDataSources(app, instructions);
   setupModels(app, instructions);
+
   setupMiddleware(app, instructions);
   setupComponents(app, instructions);
 
@@ -164,12 +165,35 @@ function setupDataSources(app, instructions) {
 function setupModels(app, instructions) {
   defineMixins(app, instructions);
   defineModels(app, instructions);
+  attachModelsToApp(app, instructions);
+  customizeModels(app, instructions);
+}
 
+function attachModelsToApp(app, instructions) {
   instructions.models.forEach(function(data) {
     // Skip base models that are not exported to the app
     if (!data.config) return;
 
+    debug('Attaching model %s to app', data.name);
     app.model(data._model, data.config);
+  });
+}
+
+function customizeModels(app, instructions) {
+  instructions.models.forEach(function(data) {
+    var model = data._model;
+
+    if (data.sourceFile) {
+      debug('Loading customization script %s', data.sourceFile);
+      var code = require(data.sourceFile);
+      if (typeof code === 'function' && model) {
+        debug('Customizing model %s', data.name);
+        code(model);
+      } else {
+        debug('Skipping model file %s - `module.exports` is not a function',
+          data.sourceFile);
+      }
+    }
   });
 }
 
@@ -212,17 +236,6 @@ function defineModels(app, instructions) {
     } else {
       debug('Creating new model %s %j', name, data.definition);
       model = registry.createModel(data.definition);
-      if (data.sourceFile) {
-        debug('Loading customization script %s', data.sourceFile);
-        var code = require(data.sourceFile);
-        if (typeof code === 'function') {
-          debug('Customizing model %s', name);
-          code(model);
-        } else {
-          debug('Skipping model file %s - `module.exports` is not a function',
-            data.sourceFile);
-        }
-      }
     }
 
     data._model = model;


bajtos:
The current design, where model scripts are run before the model is attached to an app, is intentional. I'll try to recall the arguments for that design, so that we can check if they are still valid today.

bajtos:
See https://github.com/strongloop/loopback-boot/pull/10#issuecomment-46082125

> > IIRC the previous config function's signature was `module.exports = function(app)`

> That's correct, we are aware the prototype is going to change and the `app` object will be no longer available as an injected variable.

> Here is the thing: we want to decouple the model definition (which is done at the time when there may be no app object available yet) from the model configuration in the application. This should allow us to build built-it models like `User` or `AccessToken` using the same mechanism as is being used in the LoopBack applications.

> We have discussed the implications of this change with @ritch, all uses cases boiled down to accessing other Models, which is already possible via model relations or by querying the model registry in loopback.

Here are my concerns about the change introduced by this patch:

 1. Model customization code may want to alter model settings before the model is attached, for example add extra properties or modify per-database column mappings. Are we sure that these changes can be safely made after the model was already attached?

 2. I think mixins should be defined only after model customization script was executed. This is important in case that a mixin is expected to modify properties/methods defined by the customization script.

 3. There is also a philosophical reason for not providing the app object at model setup time. In my opinion, coupling models with the application they are running could lead to models that are difficult to reuse between applications (for example a browserified loopback app and the server app; but also different server-side apps/projects).

It seems to me that this change is related to loopback-workspace, is that correct? I would personally like to eventually refactor loopback-workspace so that it becomes a set of models (plus a connector), so that consumers (e.g. strong-arc) can mount these models on their top-level loopback app, instead of mounting loopback-workspace app as a sub-app.

Thoughts?

/cc @raymondfeng

ritch:
> It seems to me that this change is related to loopback-workspace, is that correct? I would personally like to eventually refactor loopback-workspace so that it becomes a set of models (plus a connector), so that consumers (e.g. strong-arc) can mount these models on their top-level loopback app, instead of mounting loopback-workspace app as a sub-app.

I agree about the general refactoring you are proposing for the workspace.

 - a library of models, without dependency on the state of an application
 - a stateless / mostly dependency free connector (eg. shouldn't even depend on the above models)

The issue I keep running into is outlined in the code below:

```js
// my-model.js
// desired impl.
module.exports = function(MyModel) {
  var SomeOtherModel = ???;

  MyModel.myFunc = function(cb) {
    this.find(function(err, results) {
      SomeOtherModel.someOtherFunc(results, cb);
    });
  }
}

// the only full proof impl.
module.exports = function(MyModel) {
  MyModel.myFunc = function(cb) {
    var SomeOtherModel = this.app.models.SomeOtherModel;
    this.find(function(err, results) {
      SomeOtherModel.someOtherFunc(results, cb);
    });
  }
}

// inbetween proposal
module.exports = function(MyModel, registry) {
  var SomeOtherModel = registry.get('SomeOtherModel');

  MyModel.myFunc = function(cb) {
    this.find(function(err, results) {
      SomeOtherModel.someOtherFunc(results, cb);
    });
  }
}
```

bajtos:
I am wondering if there is any way how to provide a deferred value that would throw at call-time when the model does not exist in the app.

```js
// inbetween proposal
module.exports = function(MyModel, registry) {
  // don't look up the model yet, create a handle (reference) only
  var SomeOtherModel = registry.ref('SomeOtherModel');

  MyModel.myFunc = function(cb) {
    this.find(function(err, results) {
      // resolve looks up the model and throws if the model does not exist
      SomeOtherModel.resolve().someOtherFunc(results, cb);
    });
  }
}
```

Thoughts?

raymondfeng:
Calling resolve() on demand will be a big headache.

bajtos:
> I think mixins should be defined only after model customization script was executed. This is important in case that a mixin is expected to modify properties/methods defined by the customization script.

I take this back. On the second look, I think it's more consistent to run customization scripts after all declarative (JSON) configuration was processed.

bajtos:
I think we all agree that the current solution adds unnecessary friction to the development process and should be improved. While I am not entirely happy with the proposal here, I don't have any better solution either. Let's get this patch landed then and let the practical usage tell us how good/bad it is.

I am proposing to release this as a major version, as it may break assumptions made by existing applications.

@ritch the patch fails the build ATM, could you please fix it? It would be also nice to rebase this PR on top of the current master.

bajtos:
@ritch One more thing, please add a unit test that would fail without this patch in place. I think the example snippets in our comments (`MyModel` and `SomeOtherModel`) provide good scenarios to test.

BTW when the model is attached, the application object can be accessed via `MyModel.app`. Because of that, adding `app` as the second argument of the model customization function may not be worth it. `registry` makes more sense to me, as that nudges users towards using only model registry and don't use other app entities like data sources. Just an idea to consider, I don't really mind either way.

ritch:
I've run into cases where `MyModel.app` was not defined, but I cannot recall the exact circumstances.

If we are going to break BC and release a major version I would rather approach the problem again with that in mind. That will allow for a much more elegant solution.

I think the issue we keep running into is:

 - `boot()` has a rigid order for loading `ModelDefinitions`, running customization scripts, etc
 - `ModelCtor` state is mutated in various ways throughout this process
 - anything that tries to interact with the state (eg. a `ModelCtor`) has to be aware of the order
 - mixins, customization scripts, boot scripts, modules that require `app` all must be written with the load order in mind

I think it might make it a lot easier to write mixins, customization scripts, etc, if you did not have to be aware of this order at all.

One idea is to move all accessors of this stateful information into methods / functions. Examples of this are

 - `app.models.MyModel => app.model('MyModel')`
 - `app.dataSources.db => app.dataSource('db')`

This would allow us to `boot()` on the demand. If we run the `boot()` code every time the method is run, and cache some of the heavier operations (eg. loading files), we wouldn't have the weird state awareness problem above.

Instead of `boot()` running a sequence of, defining a model, running a customization script, executing mixins, etc, it would simply load the scripts and model definitions into the registry for later use. Then when something calls `app.model()` (for the first time):

 - the definition is used to create a new `ModelCtor`
 - the customization script is executed
 - mixins are executed
 - the model is attached to the specified datasource

subsequent calls to `app.model()` would just return the previously created (and cached) `ModelCtor`.

For BC we could define getters for each model that run `app.model()` and similar for `DataSource`.

rhalff:
Found myself having the same need, where the model I need a reference to is not declared as a relation.

I've build upon the third example and the below seems to work:

```es
import registry from '../../server/server'

export default (MyModel) => {
  MyModel.myFunc = function myFunc(cb) {
    const SomeOtherModel = registry.getModel('SomeOtherModel')
    this.find((err, results) => {
      SomeOtherModel.someOtherFunc(results, cb)
    })
  }
}
```
The location where I get the registry from is probably not really desired, but could be easily changed later on.

Calling `getModel()` from within the `myFunc` method ensures the model is already loaded and the call itself is rather cheap.

`findModel()` seems to do what `get()` is proposed to do, return the reference directly. 
A `get` alias wouldn't hurt though, `findModel` sounds rather expensive.

bajtos:
I'm closing this PR as it looks rather abandoned to me. @ritch feel free to reopen (or better - create a new GH issue) if you think we should continue the discussion.

mrfelton:
Personally I feel this is a major DX issue and should be addressed. I thought the direction the patch was headed was sane and still relevant.

PR #148, opened 2015-08-11
Allow middleware array merge by a key in item objects

raymondfeng:
/to @bajtos 

See https://github.com/strongloop/loopback/issues/1581

### DIFF
--- a/lib/config-loader.js
+++ b/lib/config-loader.js
@@ -1,6 +1,7 @@
 var fs = require('fs');
 var path = require('path');
 var debug = require('debug')('loopback:boot:config-loader');
+var assert = require('assert');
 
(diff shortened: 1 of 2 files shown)


bajtos:
LGTM.

The complexity of merging rules makes me wonder whether it's time to step back and try to come up with an easier solution for this?

raymondfeng:
@bajtos I agree. It's not only for middleware but for all other configurations. We need to come up a consistent convention so that the following intentions can be expressed:

1. Add a new entry
2. Merge with an existing entry
  - Add a new property
  - Update an existing property
  - Remove an existing property (set the value to null?)
3. Remove an existing entry (set the value to null?)

PR #149, opened 2015-08-13
Resolve ${var} values in middleware.json

hacksparrow:
Declaratively load **loopback.rest** using `middleware.json`:

```
{
  "routes": {
    "loopback#rest": {
      "paths": ["${restApiRoot}"]
    }
  }
}
```

Connect to strongloop/loopback-boot#75

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -316,8 +316,48 @@ function setupMiddleware(app, instructions) {
     }
     assert(typeof factory === 'function',
       'Middleware factory must be a function');
+    data.config = getUpdatedConfigObject(data.config, app);
     app.middlewareFromConfig(factory, data.config);
   });
+}
+
+function getUpdatedConfigObject(config, app) {
+
+  var DYNAMIC_CONFIG_PARAM = /\$\{(\w+)\}$/;
+
+  function getConfigVariable(param) {
+    var configVariable = param;
+    var match = configVariable.match(DYNAMIC_CONFIG_PARAM);
+    if (match) {
+      var appValue = app.get(match[1]);
+      if (appValue !== undefined) {
+        configVariable = appValue;
+      } else {
+        console.warn('%s does not resolve to a valid value. ' +
+        '"%s" must be resolvable by app.get().', param, match[1]);
+      }
+    }
+    return configVariable;
+  }
+
+  function interpolateVariables(config) {
+    var interpolated = {};
+    Object.keys(config).forEach(function(configKey) {
+      var value = config[configKey];
+      if (Array.isArray(value)) {
+        interpolated[configKey] = value.map(getConfigVariable);
+      } else if (typeof value === 'string') {
+        interpolated[configKey] = getConfigVariable(value);
+      } else if (typeof value === 'object' && Object.keys(value).length) {
+        interpolated[configKey] = interpolateVariables(value);
+      } else {
+        interpolated[configKey] = value;
+      }
+    });
+    return interpolated;
+  }
+
+  return interpolateVariables(config);
 }
 
 function setupComponents(app, instructions) {
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -292,7 +292,7 @@ describe('executor', function() {
       });
     });
 
-    describe ('for mixins', function() {
+    describe('for mixins', function() {
       var options;
       beforeEach(function() {
         appdir.writeFileSync('custom-mixins/example.js',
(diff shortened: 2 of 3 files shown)


bajtos:
Hi @hacksparrow, thank you for the pull request.

Here are my responses to your comments in https://github.com/strongloop/loopback-boot/issues/75

> are we not supporting `const` yet?

Not yet, all code must run on Node v0.10, which does not support ES2016 features.

> is there a reason why loopback#rest takes an array of paths while loopback-explorer takes a single mountPath?

loopback-boot allows users to mount the same middleware at multiple paths, this is convenient for example when configuring the same authorization middleware for multiple endpoints. That's way it's technically possible to configure `loopback#rest` for multiple endpoints, even though it's rather impractical IMO.

As for loopback-explorer, it does not use the mechanism for middleware registration, thus it provides it's own configuration option `mountPath` that expects a single string only.

Is it more clear now?

hacksparrow:
All clear, @bajtos. Thanks!

bajtos:
Now regarding the pull request: I would strongly prefer to support `${var}` syntax for all middleware and all configuration options, not only for `loopback#rest`'s  paths. See how `$!../path` is implemented for inspiration: https://github.com/strongloop/loopback-boot/blob/674a8da693ca2fe0839b5036229743cf02d1c83d/lib/compiler.js#L596-L607

Nice to have: optimise the implementation so that it either does not use lodash's cloneDeep, or at least loads only cloneDeep, not the whole lodash module. This is important because executor runs in the browser and the lodash module adds too much code to the browserified bundle.

Also please add a unit test to verify the implementation and prevent regressions in the future, see https://github.com/strongloop/loopback-boot/blob/674a8da693ca2fe0839b5036229743cf02d1c83d/test/compiler.test.js#L2139-L2209 for inspiration.

hacksparrow:
Some points for tomorrow's discussion based on the feedback:

1. `${var}` has context in case of `loopback#rest`, so it works out fine. If we want to make it generic, we will need to have some sort of context mapping object, especially if the `var` may not come from `app.get()`.
2. lodash is used extensively in `lib/compiler.js`, do we want to redo everything using our own implementation? If no, then we might as well continue using its `cloneDeep()`, else we can do it a little later.
3. Yes, unit test is required.

bajtos:
@hacksparrow I can see you added new commits, please let me know when the patch is finished and ready for the next round of review.

hacksparrow:
@bajtos not done with it yet. Need to add tests too. Will ping you once ready.

hacksparrow:
@bajtos you can review now.

hacksparrow:
@bajtos let's do the lodash thing after loopback-explorer is also loaded declaratively.

lib/executor.js
```diff
@@ -316,10 +316,49 @@ function setupMiddleware(app, instructions) {
...
+    return configVariable;
+  }
+
+  function interpolateVariables(updatedConfigObject, config) {
+    Object.keys(config).forEach(function(configKey) {
+      var value = config[configKey];
+      if (Array.isArray(value)) {
+        updatedConfigObject[configKey] = value.map(function(item) {
+          return getConfigVariable(item);
+        });
```
bajtos: Please simplify:

```js
updatedConfigObject[configKey] = value.map(getConfigVariable);
```

lib/executor.js
```diff
@@ -316,10 +316,49 @@ function setupMiddleware(app, instructions) {
...
+    var configVariable = param;
+    var match = configVariable.match(DYNAMIC_CONFIG_PARAM);
+    if (match) {
+      configVariable = app.get(match[1]);
+      assert(configVariable, param + ' does not resolve to a valid object');
+    }
+    return configVariable;
+  }
+
+  function interpolateVariables(updatedConfigObject, config) {
```
bajtos: Instead of passing in `updatedConfigObject` as a mutable argument, I personally prefer this function to be side-effect free and return the updated (interpolated) object instead.

```js
function interpolateVariables(config) {
  var updatedConfigObject = {};
  // your original code
  return updatedConfigObject;
}
```

lib/executor.js
```diff
@@ -316,10 +316,49 @@ function setupMiddleware(app, instructions) {
...
+
+  var DYNAMIC_CONFIG_PARAM = /\$\{(\w+)\}$/;
+  var updatedConfigObject = {};
+
+  function getConfigVariable(param) {
+    var configVariable = param;
+    var match = configVariable.match(DYNAMIC_CONFIG_PARAM);
+    if (match) {
+      configVariable = app.get(match[1]);
+      assert(configVariable, param + ' does not resolve to a valid object');
```
bajtos: This is a breaking change - applications that were accidentally using `{...}` values before will stop working now. Please rework the code to emit a warning only and keep the original value. Also make sure that false-y values are not rejected.

```js
if (match) {
  var appValue = app.get(match[1]);
  if (appValue !== undefined) {
    configVariable = appValue;
  } else {
    console.warn(/*...*/);
  }
}
```

bajtos:
Reviewed. The code looks mostly good, I left few minor comments to address. Please re-run `npm test` and fix coding style issues, the CI build is failing because of them.

hacksparrow:
Thanks for the feedback @bajtos. Will make the changes accordingly.

hacksparrow:
@bajtos, pushed the recommended changes.

bajtos:
The patch LGTM :+1:

@hacksparrow  Could you please squash the commits into a single one and provide a descriptive commit message following the [50/72 rule](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)?

bajtos:
Landed, thank you!

hacksparrow:
🎉

raymondfeng:
@hacksparrow Great stuff!

superkhau:
+1

PR #151, opened 2015-08-28
Resolve ${var} values in component-config.json

hacksparrow:
App variables can now be specified in component-config.json using
the ${var} format. The value of var is looked up using app.get().

This allows loopback explorer and other loopback components which
require app variables, to be loaded declaratively using
component-config.json.


bajtos:
Nice! Landed, thank you for the contribution.

hacksparrow:
👍

PR #152, opened 2015-08-31
Add config variable checks

superkhau:
- Check for config variables defined as functions
- Check for config variables that are undefined

### DIFF
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -1,3 +1,4 @@
+var async = require('async');
 var boot = require('../');
 var path = require('path');
 var loopback = require('loopback');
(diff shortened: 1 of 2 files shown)


superkhau:
@raymondfeng Review please.

test/executor.test.js
```diff
@@ -494,6 +493,37 @@ describe('executor', function() {
       });
     });
 
+    it('should not parse config variables in functions',
+        function(done) {
+      var config =  simpleMiddlewareConfig('routes', {
+        path: Function
```
raymondfeng: This is confusing. It should be something like `path: function() {}`.

raymondfeng:
Can you also add a testcase to cover an array of string/non-string values?

superkhau:
LGTM by @raymondfeng over hangout.

PR #154, opened 2015-09-08
add env postifix to boot

yorkie:
This enables the user be able to not running `server/boot/something.prod.js` in `NODE_ENV=test`, and only works in production environment.


slnode:

yorkie:
Signed

bajtos:
@slnode ok to test

bajtos:
Hi @yorkie, thank you for the pull request. I'd like to discuss the proposed implementation a bit.

I am concerned that the current proposal is a breaking change, as it changes the current behaviour which is to run `server/boot/something.prod.js` in `NODE_ENV=test`. I'd prefer to come up with a solution that preserves backwards compatibility.

Here are two possible ways I can think of:

1) Add an option (flag) that will turn on this new feature. The flag is disabled by default for backwards compatibility.
2) Change the convention for environment-specific files. Instead of adding the suffix to a file name, we can add it to a directory name instead: `server/boot.production/something.js` is run only when `NODE_ENV=production`.

Thoughts?

/cc @raymondfeng @ritch @rmg @dashby3000

yorkie:
+1 on the way two, and I will implement it if that looks good to all :)

yorkie:
Any feedback from @raymondfeng @ritch @rmg @dashby3000?

raymondfeng:
What about `server/boot/<env>/something.js`?

yorkie:
I guess @raymondfeng's proposal would be easier to implement based on current codebase :-)

superkhau:
>What about server/boot/<env>/something.js?

+1

yorkie:
But actually as @bajtos pointed out, that requires a upgrade flag, I'm quite not sure if that is friendly to loopback users, so everyone would agree on @raymondfeng's proposal?

raymondfeng:
If we agree on the `server/boot/<env>/*.js` convention, it's a new feature that won't break existing apps and no flag is needed.

superkhau:
+1 for @raymondfeng's proposal

bajtos:
I'm fine with `server/boot/<env>/*.js`. 

Regarding backwards compatibility: this change should be backwards compatible as long as we are not scanning subdirectories (and I think we are not, are we?)

bajtos:
Sorry for the accidental close, I didn't mean to.

slnode:

bajtos:
@slnode ok to test

bajtos:
@yorkie ping, what's the status of this work? Are you still keen to get it finished?

yorkie:
Sorry @bajtos I was always busying on my other business projects last week, I will must give a push/commit in this week :-)

yorkie:
@bajtos @raymondfeng Sorry for the late updates, now I have changed my commits to follow Feng's suggestion.

BTW, do loopback's documentation has documented the async boot script? I never did know that feature, and it's fucking awesome :-)

bajtos:
> BTW, do loopback's documentation has documented the async boot script? I never did know that feature

/cc @crandmck

lib/compiler.js
```diff
@@ -71,6 +71,10 @@ module.exports = function compile(options) {
 
   bootDirs.forEach(function(dir) {
     bootScripts = bootScripts.concat(findScripts(dir));
+    if (process.env.NODE_ENV) {
```
bajtos: Please use the `env` variable instead of `process.env.NODE_ENV`, see L33.

test/executor.test.js
```diff
@@ -754,6 +755,26 @@ describe('executor', function() {
...
       .expect('passport', 'initialized', done);
   });
+
+  describe('when booting with env', function() {
+    before(function() {
+      process.env.NODE_ENV = 'test';
+    });
+    after(function() {
+      process.env.NODE_ENV = undefined;
+    });
```
bajtos: Please don't manipulate the environment, pass the desired `env` value as an argument of `envAppInstructions`.

test/executor.test.js
```diff
@@ -826,3 +847,8 @@ function simpleAppInstructions() {
   fs.copySync(SIMPLE_APP, appdir.PATH);
   return boot.compile(appdir.PATH);
 }
+
+function envAppInstructions() {
+  fs.copySync(ENV_APP, appdir.PATH);
+  return boot.compile(appdir.PATH);
```
bajtos: Here you should pass the `env` value via the options object.

bajtos:
> now I have changed my commits to follow Feng's suggestion/proposal.

Cool, I like how simple the implementation is now :) I have left few comments to address, see above.

yorkie:
Thanks for this review work, Fixed up :beers:

lib/compiler.js
```diff
@@ -71,6 +71,10 @@ module.exports = function compile(options) {
 
   bootDirs.forEach(function(dir) {
     bootScripts = bootScripts.concat(findScripts(dir));
+    if (env) {
```
yorkie: But I'm unsure if we should check this existed variable, because the `env` still is able to be set to `null` or `undefined` just like:

```js
boot.compile({env: null})
```

Any thoughts?

bajtos: If `env` is `null`, then the `||` operator will pick the next value.

```
$ node
> null || 'development'
'development'
> 
```

Thus if you specify `env: null`, then `development` will be used instead.

test/executor.test.js
```diff
@@ -754,6 +755,20 @@ describe('executor', function() {
       .get('/')
       .expect('passport', 'initialized', done);
   });
+
+  describe('when booting with env', function() {
+    it('should set the `booting` flag during execution', function(done) {
+      expect(app.booting).to.be.undefined();
+      boot.execute(app, envAppInstructions(), function(err) {
+        expect(err).to.be.undefined();
```
bajtos: Please use `if(err) return done(err)`, that produces more helpful assertion error message.

test/executor.test.js
```diff
@@ -754,6 +755,20 @@ describe('executor', function() {
...
+        expect(app.booting).to.be.false();
+        expect(
+          process.bootFlags.indexOf('barLoadedInTest') === -1).to.be.false();
```
```
expect(process.bootFlags.indexOf('barLoadedInTest')).to.equal(-1);
// or even better
expect(process.bootFlags).to.not.have.member('barLoadedInTest');
```

test/fixtures/env-app/middleware/index.js
```diff
@@ -0,0 +1,8 @@
+exports.myMiddleware = function(name) {
+  return function(req, res, next) {
+    req._names = req._names || [];
+    req._names.push(name);
+    res.setHeader('names', req._names.join(','));
+    next();
+  };
+};
```
bajtos: AFAICT, this middleware is not used by your test, please remove `env-app/middleware/index.js` and also `env-app/middleware.json`.

test/fixtures/env-app/client/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+<head lang="en">
+    <meta charset="UTF-8">
+    <title>simple-app</title>
+</head>
+<body>
+<h1>simple-app</h1>
+</body>
+</html>
```
bajtos: AFAICT, this file is not used by the tests, please remove.

test/fixtures/env-app/boot/bad.txt
```diff
@@ -0,0 +1 @@
+this is not a js file!
```
bajtos: AFAICT, this file is not used by your new test, please remove.

test/fixtures/env-app/boot/bar.js
```diff
@@ -0,0 +1,8 @@
+process.bootFlags.push('barLoaded');
+module.exports = function(app, callback) {
+  process.bootFlags.push('barStarted');
+  process.nextTick(function() {
+    process.bootFlags.push('barFinished');
+    callback();
+  });
+};
```
bajtos: AFAICT, this file is not used by your new test, please remove.

test/fixtures/env-app/boot/barSync.js
```diff
@@ -0,0 +1,4 @@
+process.bootFlags.push('barSyncLoaded');
+module.exports = function(app) {
+  process.bootFlags.push('barSyncExecuted');
+};
```
bajtos: AFAICT, this file is not used by your new test, please remove.

test/fixtures/env-app/boot/booting.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function(app, cb) {
+  if (app.booting)
+    process.bootingFlagSet = true;
+
+  process.nextTick(cb);
+};
```
bajtos: AFAICT, this file is not used by your new test, please remove.

test/fixtures/env-app/boot/foo.env.js
```diff
@@ -0,0 +1 @@
+process.bootFlags.push('fooLoadedInEnv');
```
bajtos: AFAICT, this file is not used by your new test, please remove. I believe the tests is using `boot/test/bar.js` instead, isn't?

test/fixtures/env-app/boot/foo.js
```diff
@@ -0,0 +1 @@
+process.bootFlags.push('fooLoaded');
```
bajtos: AFAICT, this file is not used by your new test, please remove.

test/fixtures/env-app/boot/test/bar.js
```diff
@@ -0,0 +1,5 @@
+process.bootFlags.push('barLoadedInTest');
+module.exports = function(app, callback) {
+  process.bootFlags.push('barStartedInTest');
+  callback();
+};
```
bajtos: AFAICT, your unit-test is using only `barLoadedInTest`, it does not check `barStartedInTest`. Please remove the unused exported function.

bajtos:
> Thanks for this review work, Fixed up 

Cool. As I was taking another look, I realised your patch is adding a bunch of code/files that are not used by your new test at all. Please remove them per my comments above.

lib/compiler.js
yorkie: Ok, I'm able to remove the if statement :-)

test/fixtures/env-app/boot/foo.env.js
yorkie: yep, it's my bad, fixing now :-)

yorkie:
@bajtos Fixed, could you have a review again, thank you :-p

yorkie:
Hold on, there is something to fix in my side :-(

yorkie:
Ok, Fixed :-)

test/fixtures/env-app/middleware.json
```diff
@@ -0,0 +1,7 @@
+{
+  "initial": {
+    "../../helpers/push-name-middleware": {
+      "params": "custom-middleware"
+    }
+  }
```
bajtos: I think this entry can be removed, please correct me if I am wrong.

bajtos:
@yorkie cool, we are almost there - see one more comment above.

Could you please squash all your commits into a single one and provide a descriptive commit message following the [50/72 rule](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)?

In case you are not familiar with git history manipulation, the commands are `git rebase -i master` and then `git push -f`.

crandmck:
> BTW, do loopback's documentation has documented the async boot script? I never did know that feature

@yorkie We have this:
https://docs.strongloop.com/display/LB/Defining+boot+scripts#Definingbootscripts-Asynchronousbootscripts

Let me know how we can improve the doc.

superkhau:
@yorkie In addition to @crandmck doc's reference, we also have a boot script generator built-in. Just run `slc loopback:boot-script` and follow the prompts. ;)

test/fixtures/env-app/middleware.json
yorkie: You are right, fixing

yorkie:
@bajtos did squash now :-)

> @yorkie We have this:
https://docs.strongloop.com/display/LB/Defining+boot+scripts#Definingbootscripts-Asynchronousbootscripts

@crandmck That seems the initial boot scripts are not using async way, so I just thought I am not able to wait for the I/O in those scripts, perhaps give the generated source code:

```
module.exports = function (app /*, callback*/) {
  // TODO
}
```

If this looks good to you guys, I'd love to do this as well :-)

bajtos:
> @bajtos did squash done :-)

Lovely, the patch LGTM. I'd like to check why the dependant projects are failing on CI before landing this.

bajtos:
@slnode test please

bajtos:
> perhaps give the generated source code:

> ```
module.exports = function (app /*, callback*/) {
  // TODO
}
```

> If this looks good to you guys, I'd love to do this as well :-)

I sort of like your idea, but I am not sure what are you proposing to change. The output of `slc loopback:boot-script`? Or the boot scripts included in the default API Server template ([link](https://github.com/strongloop/loopback-workspace/tree/6b537f0657cc138430e5fe72f441e95b3b04e510/templates/components/api-server/template/server/boot))?

bajtos:
Released as `loopback-boot@2.13.0`.

yorkie:
> I sort of like your idea, but I am not sure what are you proposing to change. The output of slc loopback:boot-script? Or the boot scripts included in the default API Server template (link)?

I guess rewriting default templates is enough, BTW, does the script to generate boot script has an option to set async in our REPL?

yorkie:
> Released as loopback-boot@2.13.0

@bajtos Thanks, should I update loopback itself, either?

bajtos:
> I guess rewriting default templates is enough

Could you please contribute this change yourself? The templates for `slc loopback` are [here](https://github.com/strongloop/loopback-workspace/tree/6b537f0657cc138430e5fe72f441e95b3b04e510/templates/components/api-server/template/server/boot) and for `slc loopback:boot-script` [here](https://github.com/strongloop/generator-loopback/tree/b24c39e996385eaabe128a2a22d41c5ed899bf9b/boot-script/templates).

> BTW, does the script to generate boot script has an option to set async in our REPL?

The boot-script generator asks whether you would like to create a sync or async version, see https://github.com/strongloop/generator-loopback/blob/b24c39e996385eaabe128a2a22d41c5ed899bf9b/boot-script/index.js#L40-L55

> should I update loopback itself, either?

No need to update the loopback module if you only want the feature from this pull request.

PR #155, opened 2015-09-09
test: fix strict mode failure

rmg:
Since the value is undefined at the this point, just remove it so that
linters and strict mode don't cause the entire test suite to abort.


bajtos:
Landed, thank you!

PR #156, opened 2015-10-09
Support bluemix environment variables for host and port

bajtos:
Detect Bluemix/CouldFoundry environment and automatically apply host/port settings from the environment variables.

See https://github.com/strongloop/loopback/issues/782

/to @kraman or @sam-github please review
/cc @jasnell

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -93,6 +93,7 @@ function setHost(app, instructions) {
     process.env.npm_config_host ||
     process.env.OPENSHIFT_SLS_IP ||
     process.env.OPENSHIFT_NODEJS_IP ||
+    process.env.VCAP_APP_HOST ||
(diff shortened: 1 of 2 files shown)


rmg:
@slnode test please

sam-github:
LGTM, except that I think BlueMix should use the standard `HOST` and `PORT` env variables, not namespace its own.

I understand the reluctance of a cloud provider to have non-namespaced env vars, but in this case there is a convention worth following.

@kraman do you know Sai's github handle?

sam-github:
@bajtos since bluemix isn't going to change anything this minute, we should do this in the interim.

rmg:
FYI, this broke @svennam92's work to get apps running under strong-pm on BlueMix...

PR #158, opened 2015-10-15
Bluemix favors PORT over VCAP_APP_PORT

svennam92:
VCAP_APP_PORT is soon to be deprecated; PORT is also set by Bluemix. PORT is also generally favored over VCAP_APP_PORT in most Bluemix apps.

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -93,8 +93,8 @@ function setHost(app, instructions) {
     process.env.npm_config_host ||
     process.env.OPENSHIFT_SLS_IP ||
     process.env.OPENSHIFT_NODEJS_IP ||
-    process.env.VCAP_APP_HOST ||
     process.env.HOST ||
+    process.env.VCAP_APP_HOST ||
     instructions.config.host ||
     process.env.npm_package_config_host ||
     app.get('host');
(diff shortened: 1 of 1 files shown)


bajtos:
Hi @svennam92, thank you for the pull request. The reason why I added VCAP_APP_PORT in #156 was to fix an issue reported by a user in https://github.com/strongloop/loopback/issues/782#issuecomment-75294603. The problem was reported in February this year. Was the Bluemix's support for PORT added since then?

As for this patch, #156 was already released, removing the support for VCAP_APP_PORT would be a breaking change that we generally try to avoid. I am proposing to keep support for VCAP_APP_PORT, but use it only when PORT is not defined.

```diff
     process.env.OPENSHIFT_NODEJS_PORT,
-    process.env.VCAP_APP_PORT,
     process.env.PORT,
+    process.env.VCAP_APP_PORT,
     instructions.config.port,
```

Thoughts?

bajtos:
@svennam92 ping, is the issue behind this patch still relevant?

svennam92:
Yes, lost of track of this... The affected sample is still using a work-around.  The change you proposed is good.

bajtos:
Landed, thank you!

Issue #28, opened 2014-07-29
Expected value of app.get('env')


greaterweb:
When initializing `boot` with `options`, if `options.env` is supplied, that value does not match what `app.get('env')` returns.

The [`env` variable](https://github.com/strongloop/loopback-boot/blob/master/lib/compiler.js#L29) is set based on options provided and passed to config loader methods for app, models and datasources. It is never stored beyond the compiler and config loader initialization.

In my opinion it would be beneficial to have `app.get('env')` match what is supplied in `options.env`.

bajtos:
+1

We should consider modifying the code building the `env` variable to take into account the current value of `app.get('env')` too.

bajtos:
Fixed by #164.

greaterweb:
:+1: I'll be sure to check it out

PR #164, opened 2015-11-25
Set app env if it is supplied in options object

Amir-61:
Description: Set the app env if it is supplied in options object which solves 

connect to strongloop/loopback-boot#28

Thanks a lot!

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -23,6 +23,7 @@ module.exports = function execute(app, instructions, callback) {
   patchAppLoopback(app);
   assertLoopBackVersion(app);
 
+  setEnv(app, instructions);
   setHost(app, instructions);
   setPort(app, instructions);
   setApiRoot(app, instructions);
@@ -85,6 +86,12 @@ function assertLoopBackVersion(app) {
       RANGE);
     throw new Error(msg);
   }
+}
+
+function setEnv(app, instructions) {
+  var env = instructions.env;
+  if (env !== undefined)
+    app.set('env', env);
 }
 
 function setHost(app, instructions) {
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -362,6 +362,11 @@ describe('executor', function() {
       }));
     }
 
+    it('should apply env passed in option object', function() {
+      boot.execute(app, someInstructions({ env: 'custom_env' }));
+      expect(app.get('env')).to.equal('custom_env');
+    });
+
     it('should honor host and port', function() {
       function assertHonored(portKey, hostKey) {
         process.env[hostKey] = randomPort();
@@ -853,6 +858,9 @@ function someInstructions(values) {
       boot: []
     }
   };
+
+  if (values.env)
+    result.env = values.env;
 
   if (values.files) {
     for (var k in values.files)
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -95,6 +95,7 @@ module.exports = function compile(options) {
   // loopback modifies the data. Since we are loading the data using `require`,
   // such change affects also code that calls `require` for the same file.
   var instructions = {
+    env: env,
     config: appConfig,
     dataSources: dataSourcesConfig,
     models: modelInstructions,


Amir-61:
@slnode test please

lib/executor.js
```diff
@@ -87,6 +88,10 @@ function assertLoopBackVersion(app) {
   }
 }
 
+function setEnv(app, instructions) {
```
Amir-61: @bajtos: would you have some insights to why the value of `instructions.env` becomes undefined in test cases in `executor.test.js` script , so that's why I set the default value to 'development' here:

Also it is worth mentioning in this [test script](http://ci.strongloop.com/job/loopback-boot/577/label=ubuntu-0.12/testReport/junit/executor%20with%20middleware/json/should_parse_multiple_config_variables/) it expects `test` as opposed to `development` for the [this test line](https://github.com/strongloop/loopback-boot/blob/master/test/executor.test.js#L456)

Amir-61: @bajtos: would you have some insights to why the value of `instructions.env` becomes undefined in test cases in `executor.test.js` script , so that's why I set the default value to 'development' here:

Also it is worth mentioning in this [test script](http://ci.strongloop.com/job/loopback-boot/577/label=ubuntu-0.12/testReport/junit/executor%20with%20middleware/json/should_parse_multiple_config_variables/) it expects `test` as opposed to `development` for the [this test line](https://github.com/strongloop/loopback-boot/blob/master/test/executor.test.js#L456)

lib/executor.js
```diff
@@ -87,6 +88,10 @@ function assertLoopBackVersion(app) {
...
+  app.set('env', instructions.env || 'development');
```
Amir-61: @bajtos: would you have some insights to why the value of `instructions.env` becomes undefined in test cases in `executor.test.js` script , so that's why I set the default value to 'development' here:

Also it is worth mentioning in this [test script](http://ci.strongloop.com/job/loopback-boot/577/label=ubuntu-0.12/testReport/junit/executor%20with%20middleware/json/should_parse_multiple_config_variables/) it expects `test` as opposed to `development` for the [this test line](https://github.com/strongloop/loopback-boot/blob/master/test/executor.test.js#L456)

lib/executor.js
```diff
@@ -87,6 +88,10 @@ function assertLoopBackVersion(app) {
...
+}
+
```
Amir-61: @bajtos: would you have some insights to why the value of `instructions.env` becomes undefined in test cases in `executor.test.js` script , so that's why I set the default value to 'development' here:

Also it is worth mentioning in this [test script](http://ci.strongloop.com/job/loopback-boot/577/label=ubuntu-0.12/testReport/junit/executor%20with%20middleware/json/should_parse_multiple_config_variables/) it expects `test` as opposed to `development` for the [this test line](https://github.com/strongloop/loopback-boot/blob/master/test/executor.test.js#L456)

lib/executor.js
```diff
@@ -32,6 +32,7 @@ module.exports = function execute(app, instructions, callback) {
   setupModels(app, instructions);
   setupMiddleware(app, instructions);
   setupComponents(app, instructions);
+  setEnv(app, instructions);
```
bajtos: This should be the first operation executed, just before `setHost`. That way other setup procedures (e.g. initialisation of 3rd party components) see the correct `app.get('env')` value.

lib/executor.js
bajtos: > would you have some insights to why the value of instructions.env becomes undefined in test cases in executor.test.js script , so that's why I set the default value to development here.

Many tests use hand-written instructions created by function `someInstructions`. You should extend that function to provide default `env` when it was not passed in. See https://github.com/strongloop/loopback-boot/blob/088ca864c25e5f10d1c2277e943dd786394be986/test/executor.test.js#L845-L863

> Also it is worth mentioning in this test script it expects test as opposed to development for the this test line on Jenkins!

When not set to a different value, `app.get('env')` returns `NODE_ENV` or `development`. Unit-tests are run with `NODE_ENV=test` on our CI.

I think you need to change `someInstructions` to set `env` to `process.env.NODE_ENV || 'development'`.

----

Actually, forget about what I said above. It will be best to preserve backwards compatibility here and change `setEnv` to a no-op when `instructions.env` is not set.

```js
function setEnv(app, instructions) {
  var env = instructions.env;
  if (env === undefined) return;
  app.set('env', env);
}
```

Then you should add one more test to verify that `options.env` is propagated to `app.get('env')`, if you haven't done so already.

test/executor.test.js
```diff
@@ -362,6 +362,14 @@ describe('executor', function() {
       }));
     }
 
+    it('should apply env passed in option object', function() {
+      function assertEnv(env) {
+        boot.execute(app, someInstructions({ env: env }));
+        assert.equal(app.get('env'), env);
+      }
+      assertEnv('test');
+    });
```
bajtos: What is the benefit of wrapping the test code in a function? I think the shorter version is better.

Also we are slowly abandoning `assert` in favour of `expect` from chai, could you please use it in all new code, even if the old code around uses `assert`?

Last but not least, you should use a variable that is not set in `NODE_ENV`. At the moment, your test should be always passing on our CI, providing my understanding is correct.

```js
it('should apply env passed in option object', function() {
  boot.execute(app, someInstructions({ env: 'custom-env' }));
  expect(app.get('env')).to.equal('custom-env');
});
```

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@bajtos I applied your feedback; Thanks so much! Also, still I'm getting the following [Jenkins failure](http://ci.strongloop.com/job/loopback-example-remote/141/console) which I guess should be ignored?!

```
Warning: you have no plugins providing access control for builds, so falling back to legacy behavior of permitting any downstream builds to be triggered
Triggering a new build of update-commit-status
```

Again thanks so much for your help!

richardpringle:
connected to strongloop/loopback-boot#28

bajtos:
@slnode test please

bajtos:
@Amir-61 re: commit message

> Set the app env if it is supplied in options object which solves issue
number 28 under loopback-boot repo

Either use a link, e.g. ` Set the app env if it is supplied in options object which solves #28` or don't mention the issue at all - the issue can be found by following links from the merge commit to this pull request and then to the issue.

bajtos:
The code LGTM.

The test failures looks unrelated to this change to me, I'm going to land this PR.

bajtos:
@Amir-61 landed, thank you!

PR #174, opened 2016-02-22
executor: move "booted" and cb() to the next tick

bajtos:
Fix executor to always emit the "booted" event and call the callback
in the next tick of the event loop, regardless of whether there are any
async boot scripts.

Before this change, adding a listener for "booted" event was cumbersome:

    boot(app);
    if (app.booting)
      app.on('booted', handler);
    else
      handler();

With the fix in place, one can simply write the following:

    boot(app);
    app.on('booted', handler);

/to @raymondfeng please review

/cc @doublemarked FYI. This patch was inspired by your comment on Gitter where you were explaining how to wait for the boot process to finish.

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -43,7 +43,12 @@ module.exports = function execute(app, instructions, callback) {
     function(done) {
       enableAnonymousSwagger(app, instructions);
       done();
-    }], function(err) {
+    },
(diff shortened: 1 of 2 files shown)


raymondfeng:
LGTM

doublemarked:
@bajtos cool!

PR #177, opened 2016-03-28
Dynamic configuration

davidcheung:
connect to strongloop/loopback-boot/issues/4
### Implementation: 
This uses the original implementation of **dynamic configuration** of middleware that reads from `server/config.json`, and now takes `ENV VAR` precedence over `app.get()` 

------------
datasources will now handle dynamic configuration such as
```
{
  "mysql" : {
    "name" : "mysql_db",
    "host" : "${MYSQL_DB_HOST}",
    ...
  }
}
```
and you will be able to provide the parameter through
  ENV such as
    `MYSQL_DB_HOST=127.0.0.1 node .`
  server/config.json
```
  {
    "MYSQL_DB_HOST": "127.0.0.1"
  }
```

### Following will be implemented and addressed in a different task 

>Need to handle scenario
----------------------
- [ ] ability for dev to easily handle prefix'ed(or however its processed)  **ENV** var (bluemix/heroku/..etc) 
- [ ] any concerns for us to make a config file becoming a template?
- [ ] ability to use **cloud provided** env-var data 
### deployment platform will prefix env-vars?
`MYSQL_HOST=127.0.0.1 node .`
but in deployment env 
```
//process.env 
{
  "VCAP_MYSQL_HOST": "127.0.0.1",
.....
}
```
https://console.ng.bluemix.net/docs/services/SQLDB/index.html#vcap-free
```
{
  "sqldb": [
    {
      "name": "mydb",
      "label": "sqldb",
      "plan": "sqldb_free",
      "credentials": {
        "port": 50000,
        "db": "SQLDB",
        "username": "xxxxxxx",
        "host": "75.126.155.92",
        "hostname": "75.126.155.92",
        "jdbcurl": "jdbc:db2://75.126.155.92:50000/SQLDB",
        "uri": "db2://xxxxxxx:xxxxxxx@75.126.155.92:50000/SQLDB",
        "password": "xxxxxxx"
      }
    }
  ]
}
```
For this we will need the ability to resolve more than just a string from **env var** and **config.json** but need logic to handle objects or functions

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -171,6 +171,10 @@ function applyAppConfig(app, instructions) {
 
 function setupDataSources(app, instructions) {
   forEachKeyedObject(instructions.dataSources, function(key, obj) {
+    var opts = {
+      useEnvVars: true,
+    };
+    obj = getUpdatedConfigObject(app, obj, opts);
     app.dataSource(key, obj);
   });
 }
@@ -330,24 +334,42 @@ function setupMiddleware(app, instructions) {
     }
     assert(typeof factory === 'function',
       'Middleware factory must be a function');
-    data.config = getUpdatedConfigObject(app, data.config);
+    var opts = {
+      useEnvVars: true,
+    };
+    data.config = getUpdatedConfigObject(app, data.config, opts);
     app.middlewareFromConfig(factory, data.config);
   });
 }
 
-function getUpdatedConfigObject(app, config) {
+function getUpdatedConfigObject(app, config, opts) {
   var DYNAMIC_CONFIG_PARAM = /\$\{(\w+)\}$/;
+  var useEnvVars = opts && opts.useEnvVars;
 
   function getConfigVariable(param) {
     var configVariable = param;
     var match = configVariable.match(DYNAMIC_CONFIG_PARAM);
     if (match) {
-      var appValue = app.get(match[1]);
-      if (appValue !== undefined) {
+      var varName = match[1];
+      if (useEnvVars && process.env[varName] !== undefined) {
+        debug('Dynamic Configuration: Resolved via process.env: %s as %s',
+          process.env[varName], param);
+        configVariable = process.env[varName];
+      } else if (app.get(varName) !== undefined) {
+        debug('Dynamic Configuration: Resolved via app.get(): %s as %s',
+          app.get(varName), param);
+        var appValue = app.get(varName);
         configVariable = appValue;
       } else {
-        console.warn('%s does not resolve to a valid value. ' +
-        '"%s" must be resolvable by app.get().', param, match[1]);
+        // previously it returns the original string such as "${restApiRoot}"
+        // it will now return `undefined`, for the use case of
+        // dynamic datasources url:`undefined` to fallback to other parameters
+        configVariable = undefined;
+        console.warn('%s does not resolve to a valid value, returned as %s. ' +
+        '"%s" must be resolvable in Environment variable or by app.get().',
+          param, configVariable, varName);
+        debug('Dynamic Configuration: Cannot resolve variable for `%s`, ' +
+          'returned as %s', varName, configVariable);
       }
     }
     return configVariable;
@@ -394,7 +416,10 @@ function setupComponents(app, instructions) {
   instructions.components.forEach(function(data) {
     debug('Configuring component %j', data.sourceFile);
     var configFn = require(data.sourceFile);
-    data.config = getUpdatedConfigObject(app, data.config);
+    var opts = {
+      useEnvVars: true,
+    };
+    data.config = getUpdatedConfigObject(app, data.config, opts);
     configFn(app, data.config);
   });
 }
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -442,6 +442,10 @@ describe('executor', function() {
   });
 
   describe('with middleware.json', function() {
+    beforeEach(function() {
+      delete process.env.restApiRoot;
+    });
+
     it('should parse a simple config variable', function(done) {
       boot.execute(app, simpleMiddlewareConfig('routes',
         { path: '${restApiRoot}' }
@@ -450,6 +454,36 @@ describe('executor', function() {
       supertest(app).get('/').end(function(err, res) {
         if (err) return done(err);
         expect(res.body.path).to.equal(app.get('restApiRoot'));
+        done();
+      });
+    });
+
+    it('should parse simple config variable from env var', function(done) {
+      process.env.restApiRoot = '/url-from-env-var';
+      boot.execute(app, simpleMiddlewareConfig('routes',
+        { path: '${restApiRoot}' }
+      ));
+
+      supertest(app).get('/url-from-env-var').end(function(err, res) {
+        if (err) return done(err);
+        expect(res.body.path).to.equal('/url-from-env-var');
+        done();
+      });
+    });
+
+    it('dynamic variable from `env var` should have' +
+    ' precedence over app.get()', function(done) {
+      process.env.restApiRoot = '/url-from-env-var';
+      var bootInstructions;
+      bootInstructions = simpleMiddlewareConfig('routes',
+        { path: '${restApiRoot}' });
+      bootInstructions.config = { restApiRoot: '/url-from-config' };
+      boot.execute(app, someInstructions(bootInstructions));
+
+      supertest(app).get('/url-from-env-var').end(function(err, res) {
+        if (err) return done(err);
+        expect(app.get('restApiRoot')).to.equal('/url-from-config');
+        expect(res.body.path).to.equal('/url-from-env-var');
         done();
       });
     });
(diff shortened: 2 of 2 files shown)


lib/executor.js
```diff
@@ -171,10 +171,48 @@ function applyAppConfig(app, instructions) {
 
 function setupDataSources(app, instructions) {
   forEachKeyedObject(instructions.dataSources, function(key, obj) {
+    var opts = {
+      // do we need to limit dynamic-able keys? or everything can be dynamic
+      keys: ['host', 'url', 'port']
```
bajtos: Everything should be dynamic. I think this feature should not be tied to datasources only, it should work at least for  middleware and component configuration too.

davidcheung: Okay, then I can probably just add **env var** to the dynamic configuration implementation for middlewares, and reuse that :+1:

lib/executor.js
```diff
@@ -330,24 +332,37 @@ function setupMiddleware(app, instructions) {
...
       'Middleware factory must be a function');
-    data.config = getUpdatedConfigObject(app, data.config);
+    var useEnvVar = true;
+    data.config = getUpdatedConfigObject(app, data.config, useEnvVar);
     app.middlewareFromConfig(factory, data.config);
   });
 }
 
-function getUpdatedConfigObject(app, config) {
+function getUpdatedConfigObject(app, config, useEnvVar) {
```
bajtos: Boolean arguments are difficult to read at call-site and also difficult to extend. Please use an `options` argument instead.

```js
// usage
getUpdatedConfigObject(app, obj, { useEnvVar: true });

// impl
function getUpdatedConfigObject(app, config, options) {
  var useEnvVar = options && options.useEnvVar;
  // ...
}
```

Also please rename the variable/flag to `useEnvVars` (plural form), since there can be more than a single env variable to handle.

lib/executor.js
```diff
@@ -330,24 +332,37 @@ function setupMiddleware(app, instructions) {
...
   var DYNAMIC_CONFIG_PARAM = /\$\{(\w+)\}$/;
 
   function getConfigVariable(param) {
     var configVariable = param;
     var match = configVariable.match(DYNAMIC_CONFIG_PARAM);
     if (match) {
-      var appValue = app.get(match[1]);
-      if (appValue !== undefined) {
+      if (useEnvVar && process.env[match[1]] !== undefined) {
```
bajtos: Please save `match[1]` to a named variable (e.g. `varName`) to make the code easier to read.

lib/executor.js
```diff
@@ -330,24 +332,37 @@ function setupMiddleware(app, instructions) {
...
+        debug('Dynamic Configuration: Using app.get(server/config.json): ' +
+          param, app.get(match[1]));
+        var appValue = app.get(match[1]);
         configVariable = appValue;
       } else {
         console.warn('%s does not resolve to a valid value. ' +
-        '"%s" must be resolvable by app.get().', param, match[1]);
+        '"%s" must be resolvable in Environment variable or by app.get().',
+          param, match[1]);
+        debug('Dynamic Configuration: No valid configuration found');
```
bajtos: ditto

lib/executor.js
```diff
@@ -330,24 +332,37 @@ function setupMiddleware(app, instructions) {
...
+        configVariable = process.env[match[1]];
+      } else if (app.get(match[1]) !== undefined) {
+        debug('Dynamic Configuration: Using app.get(server/config.json): ' +
+          param, app.get(match[1]));
+        var appValue = app.get(match[1]);
         configVariable = appValue;
       } else {
         console.warn('%s does not resolve to a valid value. ' +
-        '"%s" must be resolvable by app.get().', param, match[1]);
+        '"%s" must be resolvable in Environment variable or by app.get().',
```
bajtos: Please improve the message to tell the user what value will be used and also include the parameter name & variable name.

```js
console.warn('%s does not resolve to a valid value, using "undefined" instead. ' +
  // ...
```

lib/executor.js
bajtos: Please improve the message to tell the user what value will be used and also include the parameter name & variable name.

test/executor.test.js
```diff
@@ -450,6 +450,38 @@ describe('executor', function() {
...
+        if (err) return done(err);
+        expect(res.body.path).to.equal('/url-from-env-var');
+        delete process.env.restApiRoot;
+        done();
+      });
+    });
+
+    it('dynamic variable from `env var` should have' +
+    ' precedence over app.get()', function(done) {
+      process.env.restApiRoot = '/url-from-env-var';
```
bajtos: this will leave `process.env` polluted for other unit-tests, which may lead to test failure that will be difficult to debug.

Please add a `beforeEach` hook to reset `restApiRoot` to undefined value before any test is run.

test/executor.test.js
```diff
@@ -809,6 +882,78 @@ describe('executor', function() {
     });
   });
 
+  describe('dynamic configuration for datasources.json', function() {
+
+    afterEach(function() {
+      delete process.env.DYNAMIC_HOST;
+      delete process.env.DYNAMIC_PORT;
+    });
```
bajtos: In general, it's better to clean up before the test than after the test. That way when the test is over, you can still inspect the environment in which it was running. Although I admit this is more useful when it comes to files & databases than here with env variables.

bajtos:
Looks mostly good, I left few comments to address.

Also please rebase on the current master and make sure CI builds are passing.

davidcheung:
@slnode test please

test/executor.test.js
davidcheung: @bajtos I was removing `process.env` after the test upon success, was thinking the next set of test might get polluted as I am doing it on the sub-describe. Would I need to also need to handle it after? or It's okay for next set of test to handle its own environment.

lib/executor.js
davidcheung: @bajtos I have added the `<value-used>` to the message along with the `dynamic-value` (eg. `${restApiRoot}`)
but the param(orginal key [eg.`paths` ]) is not in the function, do you think its sufficient? 
[Example of parent key not being passed](https://github.com/strongloop/loopback-boot/blob/058e0e2f56024f6e991709a156e4cf5fa2438862/lib/executor.js#L333) .


>```json
"paths": [
  "${restApiRoot}"
]
```

File: lib/executor.js [hunk 8]
```diff
@@ -330,24 +334,42 @@ function setupMiddleware(app, instructions) {
...
       } else {
-        console.warn('%s does not resolve to a valid value. ' +
-        '"%s" must be resolvable by app.get().', param, match[1]);
+        // previously it returns the original string such as "${restApiRoot}"
+        // it will now return `undefined`, for the use case of
+        // dynamic datasources url:`undefined` to fallback to other parameters
+        configVariable = undefined;
+        console.warn('%s does not resolve to a valid value, returned as %s. ' +
+        '"%s" must be resolvable in Environment variable or by app.get().',
+          param, configVariable, param);
```
bajtos: I think the last argument should be `varName` instead of `param`, shouldn't it?

lib/executor.js
bajtos: Yeah, it will do. We can always improve later.

test/executor.test.js
bajtos: That's a tough question that I don't have a good answer for. Let's use `afterEach` for these env-var cleanups then and ignore the other comment https://github.com/strongloop/loopback-boot/pull/177#discussion_r58389244

lib/executor.js
```diff
@@ -330,24 +334,42 @@ function setupMiddleware(app, instructions) {
...
+    };
+    data.config = getUpdatedConfigObject(app, data.config, opts);
     app.middlewareFromConfig(factory, data.config);
   });
 }
 
-function getUpdatedConfigObject(app, config) {
+function getUpdatedConfigObject(app, config, opts) {
   var DYNAMIC_CONFIG_PARAM = /\$\{(\w+)\}$/;
+  var useEnvVars = opts.useEnvVars || undefined;
```
bajtos: Are you sure `opts` is always set? I'd prefer to make it an option argument and check that it's set. Also `|| undefined` is not needed, it just converts falsy value like `0`, `null`, empty string or `false` to `undefined`, which is also falsy.

```js
var useEnvVars = opts && opts.useEnvVars;
```

test/executor.test.js
bajtos: Ha, I see you already switched to `beforeEach` hooks, let's keep it the way you have it. It does not matter that much.

bajtos:
@davidcheung two more comments, see above.

 Please check the commit message too, I find the `server/config.json` part confusing - how is it related to the example of running the app with an env variable set? It would be good be more verbose and explain in more detail what exactly each code snippet shows.

lib/executor.js
davidcheung: In this case 
`param` would be `${restApiRoot}`
and `varName` would be `restApiRoot`, I feel either one would make sense. Will update it

lib/executor.js
davidcheung: It definitely could be not set, good catch, fixed it.

lib/executor.js
bajtos: The new version looks better to me :+1:

bajtos:
Landed, thank you @davidcheung !

PR #181, opened 2016-04-08
[SEMVER-MAJOR] Refactor for modular and pluggable design

raymondfeng:
connect to strongloop-internal/scrum-loopback/issues/1136

The current boot process consists of three steps: 
- load 
- compile
- execute

It hard-codes logic to handle various artifact types:
- data source
- model
- mixin
- middleware
- boot script

The problems with the current design/implementation are:
- lack of extensibility 
- no clean separation of different types of artifacts

The PR proposes the following changes:
- refactor logic of processing artifacts into their own classes
- introduce `Container` as the main class for bootstrapping and build a registry of handlers during boot to organize them by a hierarchy denoted by path
- adopt `middleware` like registration and invocation
  - container.use(path, handler)
  - container.run(context, done)
- allow more phases during boot
## \- explore dependency injection

Connect to #188

**Update 16/11/16**:
  - needs review by @raymondfeng, maybe @strongloop/squad-epitome can help out?


bajtos:
Hey @raymondfeng, the proposal looks good from high-level view. I have few questions for you:

1) How do you envision supporting browserify scenario, where there are two steps run in different environments: 
 - the first step (compile + browserify) is run in Node.js on a dev machine (or build server) and creates a set of instructions for the second step,
 - and the second steps executes these instructions in a browser.

2) In the current version, bootstrapping of models has two phases: in the first phase, all models are defined in juggler's model registry. In the second phase, which is run after datasources are created, the models are attached to datasources and registered with app's RemoteObject instance to expose them via public (REST) API. Does your proposal take this into account?

3) It is important to preserve the order in which different artifacts are created. For example, to attach models to datasources, datasources must be already setup; boot scripts should be generally run at the very end, when everything else is already set up. How do you propose to define this ordering?

4)
> adopt middleware like registration and invocation
>```
container.use(path, handler)
container.run(context, done)
```

What is the purpose of this feature? Could you please provide examples where one would use this feature and how the code will look like?

5) allow more phases during boot

Again, could you please show some examples of how this would be used?

6) build a registry of handlers during boot to organize them by a hierarchy denoted by path

From your proposal, it is not clear to me what these paths will be used for. Could you please elaborate a bit more? How will these paths relate to the need of having a well-defined order in which handlers are executed?

**Comment by superkhau** [comment]
While we're at it, @davidcheung brought up a good point about config.json vs config.js inconsistencies with our configurations. We should have the new loopback-boot respect base.js files too (config.js, model-config.js, datasources.js) etc.

**Comment by pbhadauria2000** [comment]
Can you also make the DYNAMIC_CONFIG_PARAM regex to a bit flexible. Something like /(.*)\$\{(\w+)\}(.*)/ and then you can do configVariable = match(1) + configVariable + match(2)

**Comment by raymondfeng** [comment]
Connect to #188

**Comment by raymondfeng** [comment]
@pbhadauria2000 
> Can you also make the DYNAMIC_CONFIG_PARAM regex to a bit flexible. Something like /(.)\${(\w+)}(.)/ and then you can do configVariable = match(1) + configVariable + match(3)

Can you provide an example?

**Comment by pbhadauria2000** [comment]
@raymondfeng 

I would like to load sampleData files for memory datasource from different folders based on variables in config.json. Example: file: "samples\{$dataFolder}\product.json". In this example, I can keep the value of the variable "dataFolder" defined in config which could be environment specific.

File: browser.js [hunk 1]
```diff
@@ -16,7 +16,7 @@ var Bootstrapper = require('./lib/bootstrapper').Bootstrapper;
  * @header boot(app)
  */
 
-exports = module.exports = function bootBrowserApp(app, options) {
+exports = module.exports = function bootBrowserApp(app, options, callback) {
```
bajtos: Since this is for LoopBack Next, can we drop support for callbacks and switch exclusively to Promise-based APIs here in loopback-boot?

raymondfeng: I want to do this incrementally. The current implementation is ready for LoopBack 3.0. We can further clean it up for LoopBack.next if we decide to drop callback style.

bajtos: Makes sense.

bajtos:
@raymondfeng could you please comment on the questions I posted in https://github.com/strongloop/loopback-boot/pull/181#issuecomment-208767034?

lib/bootstrapper.js
```diff
@@ -0,0 +1,202 @@
...
+
+  context = context || {};
+
+  var phases = context.phases || this.phases;
+  var bootPlugins = this.getExtensions('/boot');
+  async.eachSeries(phases, function(phase, cb1) {
+    debug('Phase %s', phase);
+    async.eachSeries(bootPlugins, function(plugin, cb2) {
+      var result;
+      if (typeof plugin.handler[phase] === 'function') {
```
bajtos: Use reverse logic to make the code easier to read and reduce indentation.

```js
 if (typeof plugin.handler[phase] !== 'function') {
  debug('Skipping %s.%s', plugin.handler.name, phase);
  return cb2();
}

try {
 // ...
} catch (err) {
  // ...
}
```

lib/bootstrapper.js
```diff
@@ -0,0 +1,202 @@
...
+  options.rootDir = appConfigRootDir;
+  options.env = env;
+
+  context = context || {};
+
+  var phases = context.phases || this.phases;
+  var bootPlugins = this.getExtensions('/boot');
+  async.eachSeries(phases, function(phase, cb1) {
+    debug('Phase %s', phase);
+    async.eachSeries(bootPlugins, function(plugin, cb2) {
```
bajtos: Having to call variables like `cb1` and `cb2` is a code smell. You should use more descriptive names, e.g. `nextPhase` and `nextPlugin`.

I think it will be even better to refactor this code and move inner anonymous functions out of this method into regular functions/methods.

lib/bootstrapper.js
```diff
@@ -0,0 +1,202 @@
...
+        try {
+          if (plugin.handler[phase].length === 2) {
+            plugin.handler[phase](context, cb2);
+          } else {
+            result = plugin.handler[phase](context);
+            if (typeof Promise !== 'undefined') {
+              Promise.resolve(result).then(function(value) {
+                cb2(null, value);
+              }).catch(function(err) {
+                debug(err);
```
bajtos: Please add some context to the debug message. A stack trace alone is usual not enough information for the reader to be able to understand what's happening.

lib/bootstrapper.js
```diff
@@ -0,0 +1,202 @@
...
+                cb2(err);
+              });
+            } else {
+              cb2(null, result);
+            }
+          }
+        } catch (err) {
+          debug(err);
```
bajtos: Ditto.

lib/bundler.js
```diff
@@ -8,47 +8,51 @@ var cloneDeep = require('lodash').cloneDeep;
...
  */
-
-module.exports = function addInstructionsToBrowserify(instructions, bundler) {
-  bundleModelScripts(instructions, bundler);
-  bundleMixinScripts(instructions, bundler);
-  bundleComponentScripts(instructions, bundler);
-  bundleOtherScripts(instructions, bundler);
-  bundleInstructions(instructions, bundler);
+module.exports = function addInstructionsToBrowserify(context, bundler) {
+  // bundlePluginScripts(context, bundler);
```
bajtos: I think it's crucial to bundle plugin scripts too, otherwise we may not have the code to run during the `execute` phase in the browser. Am I missing something?

Also to keep the footprint of the browser bundle small, we need a mechanism that will allow plugin authors to split their code into two files: one providing `compile` part that will be run while browserifying, the other one providing `execute` part that will be included in the browser bundle.

lib/plugins/application.js
```diff
@@ -0,0 +1,147 @@
...
+function Application(options) {
+  PluginBase.call(this, options, 'application', 'config');
+}
+
+util.inherits(Application, PluginBase);
+
+function patchAppLoopback(app) {
+  if (app.loopback) return;
+  // app.loopback was introduced in 1.9.0
+  // patch the app object to make loopback-boot work with older versions too
```
bajtos: I think it's ok to drop support for LB 1.x apps in the new version of loopback-boot and remove `patchAppLoopback` completely.

lib/plugins/application.js
```diff
@@ -0,0 +1,147 @@
...
+      console.error(
+        'When using loopback-boot with loopback <1.9, ' +
+        'the loopback module must be available for `require(\'loopback\')`.');
+    }
+    throw err;
+  }
+}
+
+function assertLoopBackVersion(app) {
+  var RANGE = '1.x || 2.x || ^3.0.0-alpha';
```
bajtos: Ditto, let's drop support for 1.x.

lib/plugins/boot-script.js
```diff
@@ -0,0 +1,90 @@
...
+
+  bootDirs.forEach(function(dir) {
+    bootScripts = bootScripts.concat(utils.findScripts(dir));
+    var envdir = dir + '/' + options.env;
+    bootScripts = bootScripts.concat(utils.findScripts(envdir));
+  });
+
+  // de-dedup boot scripts -ERS
+  // https://github.com/strongloop/loopback-boot/issues/64
+  bootScripts = _.uniq(bootScripts);
```
bajtos: This de-duping code with the comment is repeated at least in two places. Is that intentional and needed?

package.json
```diff
@@ -21,23 +21,24 @@
   },
   "license": "MIT",
   "dependencies": {
-    "async": "~0.9.0",
-    "commondir": "0.0.1",
-    "debug": "^2.0.0",
-    "lodash": "^3.6.0",
-    "semver": "^4.1.0",
-    "toposort": "^0.2.10"
+    "async": "^0.9.2",
```
bajtos: While you are changing dependencies, could you please upgrade to the latest version? Either `^1.5.0` or `^2.0.0-rc.4`.

package.json
```diff
@@ -21,23 +21,24 @@
...
+    "bluebird": "^3.3.5",
+    "commondir": "1.0.1",
```
bajtos: Missing `^` operator, is that intentional? I mean now you require exactly version `1.0.1` and nothing else.

package.json
```diff
@@ -21,23 +21,24 @@
...
-    "coffee-script": "^1.8.0",
-    "coffeeify": "^0.7.0",
+    "browserify": "^4.2.3",
     "eslint": "^2.5.3",
     "eslint-config-loopback": "^1.0.0",
-    "fs-extra": "^0.12.0",
-    "loopback": "^2.16.3",
-    "mocha": "^1.19.0",
-    "supertest": "^0.14.0"
+    "chai": "^3.5.0",
```
bajtos: Please keep the dev-dependencies sorted. (`browserify`, `chai`, `coffee*`, `eslint*`, ...)

test/browser.multiapp.test.js
```diff
@@ -35,29 +36,42 @@ describe('browser support for multiple apps', function() {
...
+      });
     });
   });
 });
 
+function addPlugins(b) {
+  var files = fs.readdirSync(path.join(__dirname, '../lib/plugins'));
+  files.forEach(function(f) {
+    b.require('../../lib/plugins/' + f,
+      { expose: './plugins/' + path.basename(f, '.js') });
```
bajtos: I would expect that `boot.compileToBrowserify` will handle inclusion of plugin sources, in the same way as it includes model scripts and boot scripts. What are the arguments for requiring LoopBack users to explicitly bundle all plugins themselves? Wouldn't it result in everybody copying four lines above, possibly with few tweaks in file locations?

test/browser.test.js
```diff
@@ -9,13 +9,21 @@ var vm = require('vm');
 var createBrowserLikeContext = require('./helpers/browser').createContext;
 var printContextLogs = require('./helpers/browser').printContextLogs;
 
+function addPlugins(b) {
+  var files = fs.readdirSync(path.join(__dirname, '../lib/plugins'));
+  files.forEach(function(f) {
+    b.require('../../../lib/plugins/' + f,
+      { expose: './plugins/' + path.basename(f, '.js') });
+  });
+}
```
bajtos: Ha, looks like my anticipation in https://github.com/strongloop/loopback-boot/pull/181/files#r63145820 was right. Can we move this to `boot.compileToBrowserify` please?

test/executor.test.js
```diff
@@ -141,46 +150,55 @@ describe('executor', function() {
...
-        {
-          name: 'UniqueName',
-          config: { dataSource: 'db' },
-          definition: { name: 'UniqueName' },
-          sourceFile: undefined,
-        },
-      ],
-    }));
+  it('defines all models first before running the config phase',
+    function(done) {
```
bajtos: To reduce whitespace changes, we use the following indentation when fixing long lines:

```js
it('defines all models first before running the config phase',
function(done) {
  // inner code stays indented at the same level
});
```

test/fixtures/browser-app-2/app.js
```diff
@@ -2,8 +2,10 @@ var loopback = require('loopback');
 var boot = require('../../../');
 
 var app = module.exports = loopback();
+app.start = function(done) {
+  boot(app, {
```
bajtos: Interesting, why is it necessary to move `boot` to `app.start`? Are you expecting all loopback applications to start using this pattern?

bajtos:
@raymondfeng I don't see any tests for `PluginBase`, please add.

bajtos:
In `test/compiler.test.js`, you define two "expectToThrow"/"expectToNotThrow" functions that don't follow Node.js convention of passing the callback argument as the last arg.

```diff
+ function expectToThrow(err, done, options)
+ function expectToNotThrow(done, options)
```

Please fix.

```diff
+ function expectToThrow(err, options, done)
+ function expectToNotThrow(options, done)
```

Also the names don't make it clean what is expected to throw/not throw, I think `expectCompileToThrow`/`expectCompileToNotThrow` are more descriptive.

test/executor.test.js
```diff
@@ -352,41 +386,51 @@ describe('executor', function() {
...
-      assertHonored('VCAP_APP_PORT', 'VCAP_APP_HOST');
-      assertHonored('PORT', 'HOST');
+      async.eachSeries([
+        { port: 'OPENSHIFT_SLS_PORT', host: 'OPENSHIFT_NODEJS_IP' },
+        { port: 'npm_config_port', host: 'npm_config_host' },
+        { port: 'npm_package_config_port', host: 'npm_package_config_host' },
+        { port: 'OPENSHIFT_SLS_PORT', host: 'OPENSHIFT_SLS_IP' },
+        { port: 'VCAP_APP_PORT', host: 'VCAP_APP_HOST' },
+        { port: 'PORT', host: 'HOST' },
+      ], function(config, done) {
```
bajtos: We usually use `next` as the callback argument name. It makes it clear that the next item will be processed when the callback is invoked. It also prevents confusion about which `done` callback is invoked, whether the inner `done` for `async.eachSeries` or the outer `done` for the unit test.

test/executor.test.js
```diff
@@ -352,41 +386,51 @@ describe('executor', function() {
...
+        if (err) return done(err);
+        expect(app.get('env')).to.equal('custom_env');
+        done();
+      });
     });
 
-    it('should honor host and port', function() {
-      function assertHonored(portKey, hostKey) {
+    it('should honor host and port', function(done) {
+      function assertHonored(portKey, hostKey, done) {
```
bajtos: Similarly here, it may be confused there are two `done` callback. I am proposing to use `cb` or `callback` as the argument name in `assertHonored`.

Actually, even better solution is to rework this unit test into a parameterized test case.

```js
[
  { port: 'OPENSHIFT_SLS_PORT', host: 'OPENSHIFT_NODEJS_IP' },
  // ..
].forEach(function(tc) {
  it('should honor ' + tc.host + ' and ' + tc.port, function(done) {
    var hostKey = tc.host;
    var portKey = tc.key;
    // copy the code from assertHonored here
  });
});
```

That way we continue running tests for other ENV variables even after one of them was detected as not working. The output of `mocha -R spec` will also nicely describe what different ENV variables are supported.

test/executor.test.js
```diff
@@ -423,173 +473,207 @@ describe('executor', function() {
...
-            expect(res.body.path).to.be.undefined();
-            cb();
-          });
+        boot.execute(app, config, function(err) {
+          if (err) return done(err);
+
+          supertest(app)
+            .get('/')
+            .end(function(err, res) {
+              expect(err).to.be.null;
```
bajtos: `expect(err).to.be.null()` was very intentional, see [Beware of libraries that assert on property access](https://github.com/moll/js-must#asserting-on-property-access). You should either revert back to chai version specified in the current package.json, or else setup chai with [dirty-chai](https://www.npmjs.com/package/dirty-chai) plugin.

bajtos:
@raymondfeng please rebase on top of the current master.

Do we know if there are any breaking changes introduced by this pull request? (Besides those proposed in my comments.) If we are confident this is fully backwards compatible, then perhaps we can release it as a next minor version in 2.x.

Thoughts?

lib/bundler.js
raymondfeng: I was trying to do so but ran into some require resolution issues. Can you help?

raymondfeng:
I think it's still a breaking change as now the steps are truly async. Please `async@0.9.2` invokes `callback` function synchronously (not in process.nextTick or setImmediate).

test/executor.test.js
raymondfeng: Hmm, interesting. We have other modules in a similar situation too.

lib/plugins/application.js
raymondfeng: Fixed

lib/plugins/application.js
raymondfeng: Fixed

lib/plugins/boot-script.js
raymondfeng: Similar code are used for boot and plugin scripts.

raymondfeng:
@bajtos Please see my comments inline to your questions:

> 1) How do you envision supporting browserify scenario, where there are two steps run in different environments: 
 - the first step (compile + browserify) is run in Node.js on a dev machine (or build server) and creates a set of instructions for the second step,
 - and the second steps executes these instructions in a browser.

I think so. I'm not an expert in Browserify. Maybe you or @ritch can help refine the story.

> 2) In the current version, bootstrapping of models has two phases: in the first phase, all models are defined in juggler's model registry. In the second phase, which is run after datasources are created, the models are attached to datasources and registered with app's RemoteObject instance to expose them via public (REST) API. Does your proposal take this into account?

My vision is to define more runtime phases so that we can align the bootstrapping of different resources that have dependencies to multiple steps.

> 3) It is important to preserve the order in which different artifacts are created. For example, to attach models to datasources, datasources must be already setup; boot scripts should be generally run at the very end, when everything else is already set up. How do you propose to define this ordering?

We can use the same idea as for 2.

> 4)
> adopt middleware like registration and invocation
>```
bootstrapper.use(path, handler)
bootstrapper.run(context, done)
```
What is the purpose of this feature? Could you please provide examples where one would use this feature and how the code will look like?

`bootstrapper.use()' allows us to register handlers programatically. 
`bootstrapper.run()` allows us to pass in context and optionally context.phases.

> 5) allow more phases during boot
Again, could you please show some examples of how this would be used?

See:
- https://github.com/strongloop/loopback-boot/blob/feature/extensibility/lib/bootstrapper.js#L74
- https://github.com/strongloop/loopback-boot/blob/feature/extensibility/lib/bootstrapper.js#L142

We can set the initial phases using `options.phases` or call `addPhases`.

> 6) build a registry of handlers during boot to organize them by a hierarchy denoted by path
From your proposal, it is not clear to me what these paths will be used for. Could you please elaborate a bit more? How will these paths relate to the need of having a well-defined order in which handlers are executed?

It's not fully used at this point since we only have the `boot` extension point/extensions. Once we start to break down for 2, I anticipate that there will be more extension points/extensions, such as data source registry, model registry, mixin registry.

File: lib/plugins/boot-script.js (hunk 9 above)
Comment (bajtos) [inline lib/plugins/boot-script.js]: I wanted to propose extracting a shared function, but I guess it's not worth the effort in this case.

**Comment by davidcheung** [comment]
@slnode test please

File: test/fixtures/browser-app-2/app.js (hunk 16 above)
Comment (bajtos) [inline test/fixtures/browser-app-2/app.js]: @raymondfeng Can we revert this change please?

File: test/fixtures/browser-app/app.js [hunk 20]
```diff
@@ -7,4 +7,6 @@ var loopback = require('loopback');
 var boot = require('../../../');
 
 var app = module.exports = loopback();
-boot(app);
+app.start = function(done) {
+  boot(app, __dirname, done);
+};
```
bajtos: Ditto, please revert. IMO, when you `require('./test/fixtures/browser-app')`, you should get back an application that is already bootstrapped. More importantly, we need a way how to get a bootstrapped application that is not listening on any http/https port, this is commonly used when generating swagger, an AngularJS client, etc.

test/fixtures/coffee-app/app.coffee
```diff
@@ -2,4 +2,5 @@ loopback = require 'loopback'
 boot = require '../../../'
 
 module.exports = client = loopback()
-boot(client)
+client.start = (done) ->
+  boot(client, __dirname, done)
```
bajtos: And the same comment applies here to. Please revert.

bajtos:
@superkhau could you please run through this patch to check for stylistic issues? You have better eyes for that than I have.

bajtos:
I was thinking hard how to move this work forward. I am proposing the following approach.

 - Since this is a breaking change, we should start development of a 3.0 version. This means branching `master` into `2.x`, then changing the version in `package.json` in `master` to `3.0.0-alpha.1`.
 - Then we can land this pull request on master mostly as it is now. However, I would like to get addressed the remaining line comments. Ideally in this pull request, in the worst case very soon after this PR is landed. We also need to add an entry to `3.0-RELEASE-NOTES.md` explaining what are the breaking changes and how to upgrade existing LoopBack apps.
 - Because there were changes made since this patch was started, we need to forward-port them to the new `master`. I think it's ok to do this incrementally, after this pull request is landed.

Along the way, I would like to transfer the ownership/supervision reponsibility to you @raymondfeng, since you will be the person most familiar with the new code base.

Last but not least, I see it very unlikely that we will be able to make this happen in time for LoopBack 3.0 release. That should not be a problem though, because loopback-boot is fairly independent and both the current 2.x and the upcoming 3.x versions of loopback-boot can be used with both 2.x and 3.x versions of LoopBack.

@raymondfeng @davidcheung @deepakrkris thoughts?

test/fixtures/browser-app/app.js
raymondfeng: When the `app` is used as a dependency or library, it will be much more flexible to defer the boot process. It's very similar as the factory pattern. I would prefer to minimize the side effect of `require` so that we can do things such as recycling the app during test. If the `app` is the main program, we can use `if(module === require.main)` to run the boot automatically.

raymondfeng:
> I was thinking hard how to move this work forward. I am proposing the following approach.

> - Since this is a breaking change, we should start development of a 3.0 version. This means branching master into 2.x, then changing the version in package.json in master to 3.0.0-alpha.1.

+1.

> - Then we can land this pull request on master mostly as it is now. However, I would like to get addressed the remaining line comments. Ideally in this pull request, in the worst case very soon after this PR is landed. We also need to add an entry to 3.0-RELEASE-NOTES.md explaining what are the breaking changes and how to upgrade existing LoopBack apps.

I don't see intentional breaking changes. Existing applications shouldn't be impacted. We can confirm by tests.

> Because there were changes made since this patch was started, we need to forward-port them to the new master. I think it's ok to do this incrementally, after this pull request is landed.
Along the way, I would like to transfer the ownership/supervision reponsibility to you @raymondfeng, since you will be the person most familiar with the new code base.

@davidcheung will port his changes from master to this PR. I'll work with him to own it.

> Last but not least, I see it very unlikely that we will be able to make this happen in time for LoopBack 3.0 release. That should not be a problem though, because loopback-boot is fairly independent and both the current 2.x and the upcoming 3.x versions of loopback-boot can be used with both 2.x and 3.x versions of LoopBack.

It would be nice to make the PR into LoopBack 3.x. Let's consider it a best effort for now.

lib/plugins/model.js
```diff
@@ -0,0 +1,302 @@
...
+
+  var sortedNames = toposort(edges);
+
+  var instructionsByModelName = {};
+  instructions.forEach(function(inst) {
+    instructionsByModelName[inst.name] = inst;
+  });
+
+  return sortedNames
+  // convert to instructions
```
superkhau: indent

lib/utils.js
```diff
@@ -0,0 +1,329 @@
...
+        if (!onlyScriptsExportingFunction)
+          results.push(otherFile);
+        else if (onlyScriptsExportingFunction &&
+          (typeof require.extensions[otherFileExtension]) === 'function') {
+          results.push(otherFile);
+        }
+      }
+    }
+  });
+  return (results.length > 0 ? results[0] : undefined);
```
superkhau: remove ()

lib/utils.js
```diff
@@ -0,0 +1,329 @@
...
+
+function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
+  var resolvedPath = resolveAppPath(rootDir, relativePath, resolveOptions);
+  if (!resolvedPath) {
+    return false;
+  }
+  var sourceDir = path.dirname(resolvedPath);
+  var files = tryReadDir(sourceDir);
+  var fixedFile = fixFileExtension(resolvedPath, files, false);
+  return (fixedFile === undefined ? resolvedPath : fixedFile);
```
superkhau: ditto

superkhau:
@bajtos @raymondfeng Minor comments, LGTM.

test/fixtures/browser-app/app.js
bajtos: > When the app is used as a dependency or library, it will be much more flexible to defer the boot process.

I don't understand. If you defer the boot process, then you are getting an empty loopback application created by `var app = loopback()`. What is such an empty app good for?

Even if we agree that it's a good idea to defer the booting process, then we should use a different name than `start` because that one is already used in our project template for something else, see https://github.com/strongloop/loopback-workspace/blob/43bafdba46eed1ee3f4152951c847e299e967b34/templates/projects/empty-server/files/server/server.js#L8 I am proposing to use `app.setup`, `app.init` or `app.boot` instead. (Note that `app.boot` clashes with LoopBack 1.x `app.boot`, but I don't think that's a big deal.)

bajtos:
@raymondfeng 

> I don't see intentional breaking changes. Existing applications shouldn't be impacted. We can confirm by tests.

I disagree, here are at least two breaking changes:

1)
https://github.com/strongloop/loopback-boot/pull/181#issuecomment-219088502
> I think it's still a breaking change as now the steps are truly async. Please `async@0.9.2` invokes `callback` function synchronously (not in process.nextTick or setImmediate). 

2)
LoopBack versions before 1.9.0 are no longer supported
https://github.com/strongloop/loopback-boot/pull/181#discussion_r63144155
https://github.com/strongloop/loopback-boot/pull/181#discussion_r63144189

Even beside that, this is a huge change, it's effectively a complete rewrite, and I don't have enough confidence in our tests to say that we are not going to break any existing apps.

Bumping up a major version is cheap, plus I prefer "safe than sorry" approach. I think we all have been bitten by a minor/patch change in a dependency that turned out to be a breaking change, let's not do the same mistake ourselves.

On the other hand, we have the advantage of controlling the templates used for creating new projects. I am fine with changing the template used by `yo loopback` and `apic loopback` to use the new version of loopback-boot in the new projects. That way new developers coming to LoopBack will get the new version of loopback-boot, while existing projects will remain safe from possible breaking changes we did not anticipate.

> It would be nice to make the PR into LoopBack 3.x. Let's consider it a best effort for now.

Sure. Just keep in mind that we have only about a week or so to make this happen.

bajtos:
@davidcheung once https://github.com/strongloop/loopback-boot/pull/211 is landed, please add a description to 3.0-RELEASE-NOTES.md as part of this pull request (see https://github.com/strongloop/loopback-boot/pull/181#issuecomment-244730315 for starter).

davidcheung:
@slnode test please

lib/bootstrapper.js
```diff
@@ -0,0 +1,210 @@
...
+
+var assert = require('assert');
+var async = require('async');
+var utils = require('./utils');
+var path = require('path');
+var pluginLoader = require('./plugin-loader');
+var debug = require('debug')('loopback:boot:bootstrapper');
+var Promise = global.Promise || require('bluebird');
+
+module.exports = function(options) {
```
bajtos: I personally prefer to export `Bootstrapper` directly. 

```js
module.exports = Bootstrapper;
```

When I look at the code above, I see 2 places with `require('...boostrapper').Bootstrapper` and only one place calling the exported function directly.

lib/bootstrapper.js
```diff
@@ -0,0 +1,210 @@
...
+  var self = this;
+  self.builtinPlugins.forEach(function(p) {
+    var factory = require('./plugins/' + p);
+    self.use('/boot/' + p, factory(options));
+  });
+
+  try {
+    loadAndRegisterPlugins(self, options);
+  } catch (err) {
+    debug(err);
```
bajtos: Please include a message explaining what went wrong, e.g.

```js
debug('Cannot load & register plugins: %s', err.stack || err);
```

lib/bootstrapper.js
```diff
@@ -0,0 +1,210 @@
...
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+var assert = require('assert');
+var async = require('async');
+var utils = require('./utils');
+var path = require('path');
+var pluginLoader = require('./plugin-loader');
+var debug = require('debug')('loopback:boot:bootstrapper');
+var Promise = global.Promise || require('bluebird');
```
bajtos: In LoopBack 3.x, we are always using `bluebird` as our promise implementation. I am proposing to do the same in loopback-boot for consistency.

```js
var Promise = require('bluebird');
```

lib/bootstrapper.js
```diff
@@ -0,0 +1,210 @@
...
+      return done();
+    }
+    debug('Invoking %s.%s', plugin.handler.name, phase);
+    try {
+      if (plugin.handler[phase].length === 2) {
+        plugin.handler[phase](context, done);
+      } else {
+        result = plugin.handler[phase](context);
+        if (typeof Promise !== 'undefined') {
+          Promise.resolve(result).then(function(value) {
```
bajtos: Is this trying to implement support for handler functions returning a promise? Here is the version that we use elsewhere in LoopBack ([example](https://github.com/strongloop/loopback-datasource-juggler/blob/85625829e065bd2a2a9fbbe9ccf47edb75e4d063/lib/observer.js#L104-L109), it is addressing many subtle edge cases that are handled incorrectly by this code.

```js
if (result && typeof result.then === 'function') {
  result.then(
    function onPluginPhaseResolved(value) {
      done(null, value); 
    },
    function onPluginPhaseRejected(err) {
      debug('Unable to invoke %s.%s()', plugin.name, phase, err); 
      done(err); 
    });
} else {
  done(null, result);
}
```

lib/bootstrapper.js
```diff
@@ -0,0 +1,210 @@
...
+ * @param {String[]} phases An array of phase names
+ * @returns {String[]} New list of phases
+ */
+Bootstrapper.prototype.addPhases = function(phases) {
+  this.phases = utils.mergePhaseNameLists(this.phases, phases || []);
+  return this.phases;
+};
+
+function pluginIteratorFactory(context, phase) {
+  return function(plugin, done) {
```
bajtos: Please use named functions wherever possible, e.g.

```js
return function executePluginPhase(plugin, done) {
  // ...
};
```

lib/plugin-base.js
```diff
@@ -0,0 +1,337 @@
...
+  var master = ifExists(name + '.json');
+  if (!master && (ifExistsWithAnyExt(name + '.local') ||
+    ifExistsWithAnyExt(name + '.' + env))) {
+    g.warn('WARNING: Main config file "%s{{.json}}" is missing', name);
+  }
+  if (!master) return [];
+
+  var candidates = [
+    master,
+    ifExistsWithAnyExt(name + '.common'),
```
bajtos: This looks like a new feature, allowing e.g. `server/datasources.common.js`. This has been discussed in https://github.com/strongloop/loopback-boot/issues/80, different postfixes (`all` and `dynamic`) were proposed.


Let's keep this change outside of this huge refactoring please and implement it in a new pull request where it's easier to discuss details.

lib/plugin-base.js
```diff
@@ -0,0 +1,337 @@
...
+  }
+};
+
+/**
+ * Load configuration files into an array of objects.
+ * Attach non-enumerable `_filename` property to each object.
+ * @param {Array.<String>} files
+ * @returns {Array.<Object>}
+ */
+PluginBase.prototype.loadConfigFiles = function(files) {
```
bajtos: AFAICT, this is a "protected" method that should be used only by PluginBase and descendants. The convention is to prefix names of non-public methods with underscore, i.e.

```js
PluginBase.prototype._loadConfigFiles
```

lib/plugin-base.js
```diff
@@ -0,0 +1,337 @@
...
+    });
+    return config;
+  });
+};
+
+/**
+ * Merge multiple configuration objects into a single one.
+ * @param {Array.<Object>} configObjects
+ */
+PluginBase.prototype.mergeConfigurations = function(configObjects) {
```
bajtos: Ditto, should be `_mergeConfigurations`?

lib/plugin-base.js
```diff
@@ -0,0 +1,337 @@
...
+  var result = configObjects.shift() || {};
+  while (configObjects.length) {
+    var next = configObjects.shift();
+    this.merge(result, next, next._filename);
+  }
+  return result;
+};
+
+PluginBase.prototype.mergeObjects = function(target, config, keyPrefix) {
```
bajtos: `_mergeObjects`?

lib/plugin-base.js
```diff
@@ -0,0 +1,337 @@
...
+  for (var key in config) {
+    var fullKey = keyPrefix ? keyPrefix + '.' + key : key;
+    var err = this.mergeSingleItemOrProperty(target, config, key, fullKey);
+    if (err) throw err;
+  }
+  return null; // no error
+};
+
+PluginBase.prototype.mergeNamedItems = function(arr1, arr2, key) {
```
bajtos: `_mergeNamedItems`

lib/plugin-base.js
```diff
@@ -0,0 +1,337 @@
...
+      }
+    }
+    if (!found) {
+      result.push(item);
+    }
+  }
+  return result;
+};
+
+PluginBase.prototype.mergeSingleItemOrProperty =
```
bajtos: `mergeSingleItemOrProperty`

lib/plugin-base.js
```diff
@@ -0,0 +1,337 @@
...
+
+    if (newValue !== null && typeof origValue === 'object') {
+      return this.mergeObjects(origValue, newValue, fullKey);
+    }
+
+    target[key] = newValue;
+    return null; // no error
+  };
+
+PluginBase.prototype.mergeArrays = function(target, config, keyPrefix) {
```
bajtos: `_mergeArrays`

lib/plugins/application.js
```diff
@@ -0,0 +1,134 @@
...
+};
+
+function Application(options) {
+  PluginBase.call(this, options, 'application', 'config');
+}
+
+util.inherits(Application, PluginBase);
+
+function assertLoopBackVersion(app) {
+  var RANGE = '2.x || ^3.0.0-alpha';
```
bajtos: We can use `2.x || 3.x` now.

lib/plugins/boot-script.js
```diff
@@ -0,0 +1,96 @@
...
+      f.func(app, function(err) {
+        debug('Async function finished %s', f.path);
+        done(err);
+      });
+    } else {
+      debug('Starting sync function %s', f.path);
+      var error;
+      try {
+        f.func(app);
+        debug('Sync function finished %s', f.path);
```
bajtos: When the sync function fails, no debug message is printed. This is inconsistent with async functions that always print a debug log. Please fix.

lib/plugins/model.js
bajtos: +1

lib/utils.js
bajtos: +1

test/bootstrapper.test.js
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+    bootstrapper.run(context, function(err) {
+      if (err) return done(err);
+      expect(context.configurations.application).to.be.an('object');
+      expect(context.configurations.bootScripts).to.be.an('array');
+      expect(context.configurations.middleware).to.be.an('object');
+      expect(context.configurations.models).to.be.an('object');
+      expect(context.configurations.tracker).to.eql('load');
+      expect(context.instructions).to.be.undefined();
```
bajtos: The assertions above will produce unhelpful errors on failure, because it will be difficult to tell whether which property did not pass the check.

A better solution:

```js
var cfgs = context.configurations;
expect(cfg.application, 'application').to.be.an('object');
// ...
expect(context.instructions, 'instructions').to.be.undefined();
```

The same comment applies to many tests below too.

test/browser.multiapp.test.js
```diff
@@ -79,28 +83,37 @@ function browserifyTestApps(apps, next) {
...
-
-  printContextLogs(context);
+  async.each(ids, function(id, done) {
+    appsInContext[id].start(done);
+  }, function(err) {
+    printContextLogs(context);
+    done();
+  });
 
   return appsInContext;
```
bajtos: An async function that also returns a value in sync way is very weird. Please pass `appsInContext` to the callback.

```js
async.each(ids, function(id, done) {
  appsInContext[id].start(done);
}, function(err) {  
  printContextLogs(context);
  done(err, appsInContext);
});
```

test/browser.multiapp.test.js
```diff
@@ -40,29 +41,32 @@ describe('browser support for multiple apps', function() {
     browserifyTestApps(apps, function(err, bundlePath) {
       if (err) return done(err);
 
-      var bundledApps = executeBundledApps(bundlePath, apps);
-      var app1 = bundledApps.defaultApp;
-      var app2 = bundledApps.browserApp2;
+      var bundledApps = executeBundledApps(bundlePath, apps, function(err) {
```
bajtos: unhandled `err`

test/browser.test.js
```diff
@@ -50,19 +48,18 @@ describe('browser support', function() {
...
-      expect(Object.keys(app.models)).to.include('Customer');
-      expect(app.models.Customer.settings)
-        .to.have.property('_customized', 'Customer');
-
-      // configured in fixtures/browser-app/component-config.json
-      // and fixtures/browser-app/components/dummy-component.js
-      expect(app.dummyComponentOptions).to.eql({ option: 'value' });
-
-      done();
+      var app = executeBundledApp(bundlePath, function(err) {
```
bajtos: unhandled `err`

test/browser.test.js
```diff
@@ -119,18 +116,19 @@ function browserifyTestApp(options, strategy, next) {
...
   vm.runInContext(code, context, bundlePath);
   var app = vm.runInContext('require("browser-app")', context);
-
-  printContextLogs(context);
-
+  app.start(function(err) {
+    printContextLogs(context);
+    done(err);
+  });
   return app;
```
bajtos: An async function that also returns a value in sync way is very weird. Please pass `app` to the callback.

```js
app.start(function(err) {
  printContextLogs(context);
  done(err, app);
});
```

test/executor.test.js
```diff
@@ -145,46 +159,55 @@ describe('executor', function() {
...
+            name: 'UniqueName',
+            config: { dataSource: 'db' },
+            definition: { name: 'UniqueName' },
+            sourceFile: undefined,
+          },
+        ],
+      }), function(err, context) {
+        if (err) return done(err);
+        expect(app.models.Customer._modelsWhenAttached).
+        to.include('UniqueName');
```
```js
expect(app..._modelsWhenAttached)
 .to.include('UniqueName');
```

test/executor.test.js
```diff
@@ -195,106 +218,115 @@ describe('executor', function() {
...
-    function doBoot() {
-      boot.execute(app, someInstructions({ files: { boot: [file] }}));
-    }
-
-    expect(doBoot).to.throw(/Cannot find module \'doesnt-exist\'/);
+    boot.execute(app, someInstructions({ bootScripts: [file] }),
+      function(err) {
+        expect(function() {
+          if (err) throw err;
+        }).to.throw(/Cannot find module \'doesnt-exist\'/);
```
bajtos: Huh?? This is overcomplicated, just use something like the following:

```js
expect(err && err.message).to.match(/Cannot find module \'doesnt-exist\'/);
```

test/executor.test.js
```diff
@@ -195,106 +218,115 @@ describe('executor', function() {
...
         'barLoaded',
         'barSyncLoaded',
         'fooLoaded',
         'barStarted',
+        'barFinished',
+        'barSyncExecuted',
       ]);
-
-      // bar finished happens in the next tick
-      // barSync executed after bar finished
```
bajtos: I think this change needs to be clearly explained in `3.0-RELEASE-NOTES.md`.

test/fixtures/browser-app-2/app.js
bajtos: @davidcheung please revert this change.

test/fixtures/browser-app/app.js
bajtos: @davidcheung please revert this change and let's discuss how to achieve Raymond's requirement later, after this pull request is landed. That way the discussion can be focused on this single feature and does not get lost in this huge change-set.

davidcheung:
@bajtos can you PTAL, the changes are mostly logically separated into their own commit, but probably still quite difficult to navigate due to size of PR :(

davidcheung:
>I think it would be worthwhile to describe the new features in 3.0-RELEASE-NOTES too (the modular design and how to build 3rd party boot plugins). In my experience, the process of writing the documentation usually discover missing pieces and/or rough edges. However, I am fine with leaving the description of the new features out of scope of this pull request.

i have only included one change in the `release-note.md`, will open another issue to document the modular design, and how to create your own plugin.

lib/plugins/model.js
```diff
@@ -116,15 +116,15 @@ function sortByInheritance(instructions) {
       return [getBaseModelName(inst.definition), inst.name];
     });
 
-  var sortedNames = toposort(edges);
+  var _sortedNames = toposort(edges);
```
bajtos: We usually don't prefix variable names with `_`.

lib/utils.js
```diff
@@ -157,8 +157,8 @@ function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
   }
   var sourceDir = path.dirname(resolvedPath);
   var files = tryReadDir(sourceDir);
-  var fixedFile = fixFileExtension(resolvedPath, files, false);
-  return (fixedFile === undefined ? resolvedPath : fixedFile);
+  var _fixedFile = fixFileExtension(resolvedPath, files, false);
```
bajtos: We usually don't prefix variable names with `_`.

bajtos:
@davidcheung also please don't forget to squash the commits before landing!

PR #183, opened 2016-04-11
parse config: should ignore null values

### LINKED ISSUE #182: TypeError: Cannot convert undefined or null to object
Since #177 my application throws the error:
```
TypeError: Cannot convert undefined or null to object
/app/api/node_modules/loopback-boot/lib/executor.js in null.<anonymous> at line 403
```

It is because I parse myself env variables for datasource config. The results of the parsing may contains `null` values but:
```js
typeof null === 'object'
```
And so:
```js
Object.keys(null)
```
throws: **TypeError: Cannot convert undefined or null to object**

The `null` values were things like `port` etc...
I updated my application for giving default port and so avoid null value but it should certainly be fixed.

LoicMahieu:
Fix #182

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -400,6 +400,8 @@ function getUpdatedConfigObject(app, config, opts) {
         interpolated[configKey] = value.map(interpolateVariables);
       } else if (typeof value === 'string') {
         interpolated[configKey] = getConfigVariable(value);
+      } else if (value === null) {
+        interpolated[configKey] = value;
       } else if (typeof value === 'object' && Object.keys(value).length) {
         interpolated[configKey] = interpolateVariables(value);
       } else {
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -543,6 +543,23 @@ describe('executor', function() {
       });
     });
 
+    it('should parse config variables with null values', function(done) {
+      boot.execute(app, simpleMiddlewareConfig('routes',
+        { nested: { info: { path: '${restApiRoot}', some: null }}}
+      ));
+
+      supertest(app).get('/').end(function(err, res) {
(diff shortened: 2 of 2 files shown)


slnode:

davidcheung:
@slnode ok to test

lib/executor.js
```diff
@@ -400,7 +400,7 @@ function getUpdatedConfigObject(app, config, opts) {
         interpolated[configKey] = value.map(interpolateVariables);
       } else if (typeof value === 'string') {
         interpolated[configKey] = getConfigVariable(value);
-      } else if (typeof value === 'object' && Object.keys(value).length) {
+      } else if (typeof value === 'object' && value !== null && Object.keys(value).length) {
```
davidcheung: Please wrap line so it doesn't exceed 80 length.

test/executor.test.js
```diff
@@ -543,6 +543,23 @@ describe('executor', function() {
...
+      boot.execute(app, simpleMiddlewareConfig('routes',
+        { nested: { info: { path: '${restApiRoot}', some: null }}}
+      ));
+
+      supertest(app).get('/').end(function(err, res) {
+        if (err) return done(err);
+        expect(res.body.nested).to.eql({
+          info: {
+            path: app.get('restApiRoot'),
+            some: null
```
davidcheung: please add comma-dangle so eslint passes

davidcheung:
@LoicMahieu Thanks for the patch! 
@bajtos PTAL

LoicMahieu:
Rebased!

bajtos:
@LoicMahieu landed, thank you for the contribution! I have released your patch in `loopback-boot@2.18.1`.

PR #185, opened 2016-04-14
Add flag var lazyConnect to ds config

jannyHou:
Connect to strongloop/loopback#2242
Add flag var lazyConnect to ds config

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -175,6 +175,11 @@ function setupDataSources(app, instructions) {
       useEnvVars: true,
     };
     obj = getUpdatedConfigObject(app, obj, opts);
+    var lazyConnect = process.env.LB_LAZYCONNECT_DATASOURCES;
+    if (lazyConnect) {
+      obj.lazyConnect =
+        lazyConnect === 'false' || lazyConnect === '0' ? false : true;
+    }
     app.dataSource(key, obj);
   });
 }
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -906,6 +906,62 @@ describe('executor', function() {
     });
   });
 
+  describe('when booting with lazy connect', function() {
+    var SAMPLE_INSTRUCTION = someInstructions({
+      dataSources: {
+        lazyConnector: {
+          connector: 'testLazyConnect',
+          name: 'lazyConnector',
+        },
+      },
+    });
+    var connectTriggered = true;
+
+    beforeEach(function() {
+      app.connector('testLazyConnect', {
+        initialize: function(dataSource, callback) {
+          if (dataSource.settings.lazyConnect) {
+            connectTriggered = false;
+          } else {
+            connectTriggered = true;
+          }
+        },
+      });
+    });
+
+    it('should trigger connect with ENV undefined', function(done) {
+      delete process.env.LB_LAZYCONNECT_DATASOURCES;
+      boot.execute(app, SAMPLE_INSTRUCTION, function() {
+        expect(connectTriggered).to.equal(true);
+        done();
+      });
+    });
+
+    it('should not trigger connect with ENV true', function(done) {
+      process.env.LB_LAZYCONNECT_DATASOURCES = 'true';
+      boot.execute(app, SAMPLE_INSTRUCTION, function() {
+        expect(connectTriggered).to.equal(false);
+        done();
+      });
+    });
+
+    it('should trigger connect with ENV false', function(done) {
+      process.env.LB_LAZYCONNECT_DATASOURCES = 'false';
+      boot.execute(app, SAMPLE_INSTRUCTION, function() {
+        expect(connectTriggered).to.equal(true);
+        done();
+      });
+    });
+
+    it('should trigger connect with ENV 0', function(done) {
+      process.env.LB_LAZYCONNECT_DATASOURCES = '0';
+      boot.execute(app, SAMPLE_INSTRUCTION, function() {
+        expect(connectTriggered).to.equal(true);
+        done();
+      });
+    });
+  });
+
   describe('dynamic configuration for datasources.json', function() {
     beforeEach(function() {
       delete process.env.DYNAMIC_HOST;


lib/executor.js
```diff
@@ -175,6 +175,7 @@ function setupDataSources(app, instructions) {
       useEnvVars: true,
     };
     obj = getUpdatedConfigObject(app, obj, opts);
+    obj.lazyConnect = true;
```
raymondfeng: Now we defer `connect` to a later phase. When will `connect` be triggered?

jannyHou: I tried with local db, I think the `connect` will be triggered when data query happens, like hitting endpoint from Explorer. And also be triggered when [`datasource.ping()`](https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L1131) is called.

jannyHou: @raymondfeng I think my previous answer is for "when connect will be called" but not "when connect will be triggered in initialize", so for example if user create a datasource from boot script, it will be triggered, since `lazyConnect` is undefined.
```
var myDS = new DataSource(dsName, dsConfig);
// will trigger connect in initialize
```

raymondfeng: I think `boot` should trigger `connect` by default. We can set an option to disable it so that swagger generation doesn't depend on live database connections.

lib/executor.js
```diff
@@ -175,6 +175,7 @@ function setupDataSources(app, instructions) {
       useEnvVars: true,
     };
     obj = getUpdatedConfigObject(app, obj, opts);
+    obj.lazyConnect = process.env.LB_LAZYCONNECT_DATASOURCES;
```
bajtos: When the env var is not set, this ads `lazyConnect` variable with `undefined` value. I think this may surprise users when debugging. Also if `lazyConnect` was set in `datasources.json`, then you will re-set it back to whatever is provided by the env variable.

I am proposing a safer approach: 1) change the value only when the ENV var is set and also 2) handle "false" and "0" values as false

```js
var lazyConnect = process.env.LB_LAZYCONNECT_DATASOURCES;
if (lazyConnect) {
  obj.lazyConnect = lazyConnect === 'false' || lazyConnect === '0' ? false : true;
}
```

bajtos:
@jannyHou please add unit-tests to verify the implementation works as intended.

lib/executor.js
jannyHou: makes sense. I should take care of `undefined`. Thanks!

test/executor.test.js
```diff
@@ -906,6 +906,43 @@ describe('executor', function() {
     });
   });
 
+  describe('when booting with lazy connect', function() {
+    it.only('should report error', function(done) {
```
bajtos: Forgotten `it.only`, please fix to `it`.

test/executor.test.js
```diff
@@ -906,6 +906,43 @@ describe('executor', function() {
...
+
+    it('should not report error', function(done) {
+      process.env.LB_LAZYCONNECT_DATASOURCES = true;
+      var datasource = {
+        mongodbDS: {
+          host: 'invalid-host',
+          port: 80,
+          connector: "mongodb",
+          name: "mongoDS",
+        },
```
bajtos: Since this configuration is same for both tests, please refactor it into a shared variable, e.g. `DATASOURCE_THROWING_CONNECTION_ERROR`

bajtos:
IMO, depending on loopback-connector-mongodb will make the tests slow and brittle. I am proposing to create a custom connector used by tests only, this connector should be very simple - throw when `lazyConnect` is not set, do no-op otherwise.

bajtos:
You can register a custom connector by calling `app.connector(name, connectorObject)`. A mock-up implementation which may not work OOTB:

```js
// this can be configured in a beforeEach hook
app.connector('throwOnConnect', {
  initialize: function(dataSource, callback) {
    if (dataSource.settings.lazyConnect) callback();
    else callback(new Error('CANNOT CONNECT'));
  }
});

// and then use the connector in tests:
app.boot(someInstructions({ dataSources: {
  'db': { connector: 'throwOnConnect', lazyConnect: false }
}});
```

jannyHou:
@bajtos I create a custom connector for test, and simply throw new error in `initialize` to avoid executing callback function [`postInit`](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/datasource.js#L315). 
If we trigger `postInit`, it will always check connector instance valid before check connection, giving an error:
> Connector is not defined correctly: it should create `connector` member of dataSource'

The error emitted in `postInit` is already checked in each connector's test case, so I think here we only need to make sure that the flag ENV is passed to connector properly.

test/executor.test.js
```diff
@@ -906,6 +906,50 @@ describe('executor', function() {
     });
   });
 
+  describe('when booting with lazy connect', function() {
+    var DATASOURCE_THROWING_CONNECTION_ERROR = {
+      lazyConnector: {
+        host: 'invalid-host',
+        port: 80,
```
bajtos: I believe `host` and `port` are not needed by tests, please remove.

bajtos:
I agree it should be enough to test that `LB_LAZYCONNECT_DATASOURCES` is converted into `lazyConnect` setting. I think throwing an exception is an overkill for that, you can set a bool variable instead.

```js
    var lazyConnect;
    beforeEach(function() {
      app.connector('reportLazyConnect', {
        initialize: function(dataSource, callback) {
          lazyConnect = dataSource.settings.lazyConnect; 
        },
      });
    });

    var REPORT_LAZY_CONNECT_INSTRUCTIONS = someInstructions({ ... });

  it('does not set lazyConnect by default', function() {
    delete process.env.LB_LAZYCONNECT_DATASOURCES;
    boot.execute(app, REPORT_LAZY_CONNECT_INSTRUCTIONS);
    expect(lazyConnect).to.equal(undefined);
  });

  it('sets lazyConnect:true when LB_LAZYCONNECT_DATASOURCES=true', function() {
    process.env.LB_LAZYCONNECT_DATASOURCES = 'true';
    boot.execute(app, REPORT_LAZY_CONNECT_INSTRUCTIONS);
    expect(lazyConnect).to.equal(undefined);
  });

  // etc, check for cases when lazyConnect is set to "false"
});
```

jannyHou:
I copied your question from slack:
> so what would happen if we callback and let `postInit` run?

It will execute `_setupConnector()` which makes things complex. I am afraid we need to create a connector object to get it pass, instead of throwing `undefined connector` before it check connection. I need to investigate what to be included in the connector object if it's worth. 

```js
DataSource.prototype._setupConnector = function() {
  this.connector = this.connector || this.adapter; // The legacy JugglingDB adapter will set up `adapter` property
  this.adapter = this.connector; // Keep the adapter as an alias to connector
  if (this.connector) {
    if (!this.connector.dataSource) {
      // Set up the dataSource if the connector doesn't do so
      this.connector.dataSource = this;
    }
    var dataSource = this;
    this.connector.log = function(query, start) {
      dataSource.log(query, start);
    };

    this.connector.logger = function(query) {
      var t1 = Date.now();
      var log = this.log;
      return function(q) {
        log(q || query, t1);
      };
    };
    // Configure the connector instance to mix in observer functions
    jutil.mixin(this.connector, OberserverMixin);
  }
};
```

test/executor.test.js
```diff
@@ -906,6 +906,43 @@ describe('executor', function() {
     });
   });
 
+  describe.only('when booting with lazy connect', function() {
```
bajtos: Forgotten `only`, please remove. BTW if you want to run a subset of tests only, you can use mocha's command-line flag `-g`: 

```bash
$ mocha -g "lazy connect"
```

That way you don't need to change code.

test/executor.test.js
```diff
@@ -906,6 +906,43 @@ describe('executor', function() {
...
+    var DATASOURCE_THROWING_CONNECTION_ERROR = {
```
bajtos: This datasource is no longer throwing errors, please rename.

test/executor.test.js
```diff
@@ -906,6 +906,43 @@ describe('executor', function() {
...
+            connectTriggered = false;
+          }
+        },
+      });
+    });
+
+    it('should trigger connect', function(done) {
+      boot.execute(app, someInstructions({
+        dataSources: DATASOURCE_THROWING_CONNECTION_ERROR,
+      }));
```
bajtos: To avoid repetition, please extract `someInstructions({ dataSources: DATASOURCE_THROWING_CONNECTION_ERROR }); ` into a shared constant. Then you can remove `DATASOURCE_THROWING_CONNECTION_ERROR` entirely (inline it in the new constant).

test/executor.test.js
```diff
@@ -906,6 +906,43 @@ describe('executor', function() {
...
+      expect(connectTriggered).to.be.true;
+      done();
```
bajtos: Is it a good idea to rely on the fact that loopback-boot initializes datasources in the first tick? I think it will be safer to wait until loopback-boot is done:

```js
boot.execute(..., function(err) {
  if (err) return done(err);
  expect(connectTriggered)...
  done();
});
```

Thoughts?

Alternatively, if you prefer to stick with the sync way, then remove `done` from the tests as it's not needed:

```js
it('...', function() {
  boot.execute(...);
  expect(connectTriggered)...;
});

File: test/executor.test.js [hunk 10]
```diff
@@ -906,6 +906,43 @@ describe('executor', function() {
...
+    it('should not trigger connect', function(done) {
+      process.env.LB_LAZYCONNECT_DATASOURCES = true;
+      boot.execute(app, someInstructions({
+        dataSources: DATASOURCE_THROWING_CONNECTION_ERROR,
+      }));
+      expect(connectTriggered).to.be.false;
+      done();
+    });
+  });
+
```
bajtos: Please add unit-tests for the following cases too - notice the value is always a string!

 - `LB_LAZYCONNECT_DATASOURCES='false'`
 - `LB_LAZYCONNECT_DATASOURCES='0'`

test/executor.test.js
```diff
@@ -906,6 +906,43 @@ describe('executor', function() {
...
+    it('should trigger connect', function(done) {
+      boot.execute(app, someInstructions({
+        dataSources: DATASOURCE_THROWING_CONNECTION_ERROR,
+      }));
+      expect(connectTriggered).to.be.true;
+      done();
+    });
+
+    it('should not trigger connect', function(done) {
+      process.env.LB_LAZYCONNECT_DATASOURCES = true;
```
bajtos: I believe that environment variables have always string type, i.e. this line should be

```js
process.env.LB_LAZYCONNECT_DATASOURCES = `true`;
```

bajtos:
> It will execute _setupConnector() which makes things complex. I am afraid we need to create a connector object to get it pass, otherwise it will throw undefined connector before checking connection. I need to investigate what to be included in the connector object if it's worth.

Let's try the following and if you cannot get it working in under 30 minutes then leave the code as it is now and move on.

```js
// somewhere on the top of the file, assuming var loopback = require('loopback')
var Connector = loopback.Connector;

// in your initialize function
initialize: function(dataSource, callback) {
  dataSource.connector = new Connector(dataSource.options);

  if (dataSource.settings.lazyConnect) {
    connectTriggered = false;
  }
}
```

test/executor.test.js
jannyHou: oops I didn't realize that, sync way is super unreliable. I will do the check in callback function.

jannyHou:
@bajtos i add the other two test cases. And i think let's just don't trigger `postInit` this time:( Thanks for understanding. I will look into it if i have plenty time this sprint or in the near future. Probably still need to touch it when i learn oracle with raymond's pr.

test/executor.test.js
```diff
@@ -906,6 +906,62 @@ describe('executor', function() {
     });
   });
 
+  describe('when booting with lazy connect', function() {
+    var sampleInstruction = someInstructions({
```
bajtos: Please rename to indicate that this should be treated as an immutable constant, e.g. `SAMPLE_INSTRUCTIONS`.

test/executor.test.js
```diff
@@ -906,6 +906,62 @@ describe('executor', function() {
...
+      dataSources: {
+        lazyConnector: {
+          connector: 'testLazyConnect',
+          name: 'lazyConnector',
+        },
+      },
+    });
+    var connectTriggered = true;
```
bajtos: Nitpick: the value is always reset in `beforeEach` hook below, there is no need to assign a value here.

```js
var connectTriggered;
beforeEach(function() {
  connectTriggered = true;
  // ...
});
```

test/executor.test.js
```diff
@@ -906,6 +906,62 @@ describe('executor', function() {
...
+            connectTriggered = false;
+          }
+        },
+      });
+    });
+
+    it('should trigger connect with ENV undefined', function(done) {
+      delete process.env.LB_LAZYCONNECT_DATASOURCES;
+      boot.execute(app, sampleInstruction, function() {
+        expect(connectTriggered).to.be.true;
```
bajtos: [Beware of libraries that assert on property access](https://github.com/moll/js-must#asserting-on-property-access). We should start using dirty-chai, but that's out of scope of this patch. As a workaround, use this please:

```js
expect(connectTriggered).to.equal(true);
```

See also https://github.com/strongloop/eslint-config-loopback/pull/6

test/executor.test.js
```diff
@@ -906,6 +906,62 @@ describe('executor', function() {
...
+        done();
+      });
+    });
+
+    it('should not trigger connect with ENV true', function(done) {
+      process.env.LB_LAZYCONNECT_DATASOURCES = 'true';
+      boot.execute(app, sampleInstruction, function() {
+        expect(connectTriggered).to.be.false;
```
bajtos: ditto

test/executor.test.js
```diff
@@ -906,6 +906,62 @@ describe('executor', function() {
...
+        done();
+      });
+    });
+
+    it('should trigger connect with ENV false', function(done) {
+      process.env.LB_LAZYCONNECT_DATASOURCES = 'false';
+      boot.execute(app, sampleInstruction, function() {
+        expect(connectTriggered).to.be.true;
```
bajtos: ditto

test/executor.test.js
```diff
@@ -906,6 +906,62 @@ describe('executor', function() {
...
+        done();
+      });
+    });
+
+    it('should trigger connect with ENV 0', function(done) {
+      process.env.LB_LAZYCONNECT_DATASOURCES = '0';
+      boot.execute(app, sampleInstruction, function() {
+        expect(connectTriggered).to.be.true;
```
bajtos: ditto

bajtos:
Few more details to improve, the patch LGTM otherwise. No more reviews are needed.

test/executor.test.js
jannyHou: @bajtos actually the variable will not reset lol. it's defined outside the hook, and stay the same value as previous test without reset.

But i think a more reasonable logic is 
```js
if (dataSource.settings.lazyConnect) {
     connectTriggered = false;
} else {
     connectTriggered = true;
}
```
then there is no need to reset it at the beginning.

bajtos: Please do reset it at the beginning. If it happens that the code in your custom connector does not get called by some weird error that is swallowed, or perhaps because the connector API changed and our mocked connector is no longer valid, then test will see the result of the previous call, which is not right.

IMO, the most correct solution is to:
 - reset the variable in `beforeEach` hook to `undefined` or `null`
 - inside the connector, set the variable to true/false depending on the value of `lazyConnect`, as you shown in the code snippet above

jannyHou: Agreed, will fix it with a reset

PR #194, opened 2016-06-29
Stop caching config files

bajtos:
This should fix the CI failures @loay is observing in https://github.com/strongloop/loopback-workspace/pull/294 and myself in https://github.com/strongloop/loopback-workspace/pull/298

@raymondfeng please review

cc @davidcheung @0candy @gunjpan @richardpringle

### DIFF
--- a/lib/config-loader.js
+++ b/lib/config-loader.js
@@ -3,6 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+var cloneDeep = require('lodash').cloneDeep;
 var fs = require('fs');
 var path = require('path');
 var debug = require('debug')('loopback:boot:config-loader');
@@ -125,11 +126,12 @@ function findConfigFiles(appRootDir, env, name) {
  */
 function loadConfigFiles(files) {
   return files.map(function(f) {
-    var config = require(f);
+    var config = cloneDeep(require(f));
     Object.defineProperty(config, '_filename', {
       enumerable: false,
       value: f,
     });
+    debug('loaded config file %s: %j', f, config);
     return config;
   });
 }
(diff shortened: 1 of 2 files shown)


lib/config-loader.js
```diff
@@ -3,6 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+var cloneDeep = require('lodash').cloneDeep;
```
richardpringle: lodash has a lot of useful utility functions, might make more sense to require it on its own so that someone can use other lodash functions without changing this line.

davidcheung: another note on lodash, we had [previous goals to shrink bundle size](https://github.com/strongloop/loopback/issues/989), would this be going against that, 
but CI being more critical, another option would be: we are using [`traverse.clone()`](https://www.npmjs.com/package/traverse) in [datasource-juggler](https://github.com/strongloop/loopback-datasource-juggler/blob/6bc44418509a7269cdd2e862d806a091ec173003/lib/datasource.js#L427)

bajtos: > lodash has a lot of useful utility functions, might make more sense to require it on its own so that someone can use other lodash functions without changing this line.

I am just following the style seen in existing code, see e.g. https://github.com/strongloop/loopback-boot/blob/296c38568cc13a7d4fc80c1aec313eef801d06ae/lib/compiler.js#L7

> another note on lodash, we had previous goals to shrink bundle size, would this be going against that, 

Bundle size is important, thank you for raising this concern. The browser bundle contains only the source code from `lib/executor.js`, config loader is executed during compilation, therefore it's ok to use lodash here.

bajtos: @davidcheung @richardpringle PTAL :point_up:

richardpringle: I stand by what I said but maybe that's just my preference.

It shouldn't stop this PR from being landed though.

richardpringle:
LGTM

davidcheung:
LGTM2

bajtos:
Thank you for the review, I have landed the patch.

PR #195, opened 2016-07-01
test: fix security warning

bajtos:
Add a simple implementation of getRandomValues to the browser context to get rid of the following warning:

    [SECURITY] node-uuid: crypto not usable, falling back to insecure
    Math.random()

@davidcheung please review
cc @0candy @gunjpan @richardpringle

### DIFF
--- a/test/helpers/browser.js
+++ b/test/helpers/browser.js
@@ -28,6 +28,17 @@ function createContext() {
     // used by crypto-browserify & friends
     Int32Array: Int32Array,
     DataView: DataView,
+    crypto: {
+      getRandomValues: function(typedArray) {
+        var randomBuffer = require('crypto').randomBytes(typedArray.length);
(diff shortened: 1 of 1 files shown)


davidcheung:
@bajtos LGTM 

We are already skipping the browserify test in loopback-core, so I think there shouldn't be problems.
https://github.com/strongloop/loopback/commit/34eb0e1a4577739d5ec8221238dcb2b0c92a8f87

might even be worth it to port this to the loopback-core to replace skipped test?

bajtos:
> We are already skipping the browserify test in loopback-core, so I think there shouldn't be problems.
strongloop/loopback@34eb0e1

> might even be worth it to port this to the loopback-core to replace skipped test?

Possibly. I'd rather check whether the bug was fixed since the time we left the comment in LoopBack.

PR #203, opened 2016-08-09
Adding UMD support to help allow Typescript

tonicblue:
Relating to [issue 202](https://github.com/strongloop/loopback-boot/issues/202)

Have included tests and added 2 lines to `.eslintignore` to ignore generated code.

Signed-off-by: Matthew Wilkes matthew.wilkes@scvo.org.uk

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -249,7 +249,7 @@ function defineModels(app, instructions) {
           debug('Customizing model %s', name);
           code(model);
         // If code is a default UMD module use it
-        } else if (code.hasOwnProperty('default')) {
+        } else if (code.__esModule && code.hasOwnProperty('default')) {
           if (typeof code.default === 'function') {
             debug('Customizing model %s', name);
             code.default(model);
@@ -302,7 +302,7 @@ function runScripts(app, list, callback) {
           func: exports,
         });
       // If the file contains a default UMD module, load it
-      } else if (exports.hasOwnProperty('default')) {
+      } else if (exports.__esModule && exports.hasOwnProperty('default')) {
         if (typeof exports.default === 'function') {
           debug('Exported function detected %s', filepath);
           functions.push({


slnode:

slnode:

slnode:

lib/executor.js
```diff
@@ -248,6 +248,12 @@ function defineModels(app, instructions) {
         if (typeof code === 'function') {
           debug('Customizing model %s', name);
           code(model);
+        // If code is a default UMD module use it
+        } else if (code.hasOwnProperty('default')) {
```
bajtos: Can we make this check more robust by also checking for the presence of `code. __esModule`?

lib/executor.js
```diff
@@ -295,6 +301,15 @@ function runScripts(app, list, callback) {
           path: filepath,
           func: exports,
         });
+      // If the file contains a default UMD module, load it
+      } else if (exports.hasOwnProperty('default')) {
```
bajtos: ditto

lib/executor.js
tonicblue: Makes perfect sense. Just committed an update. 

Cheers

bajtos:
Hi @tonicblue, sorry for the delay. The patch LGTM. However, I am little bit worried about upgrading the test fixture files when TypeScript and/or ES6 changes how the code is generated and modules exported.

I am proposing to add the original `ts` files you used to produce `test/fixtures/simple-app/boot/umd.js` and `test/fixtures/umd-app/models/customer.js`, and leave a comment in each of those TS files explaining how to convert them to JS files.

Thoughts?

bajtos:
@davidcheung could you please take this patch over from me, since you are most familiar with the new refactored codebase now?

slnode:

bajtos:
@raymondfeng @ritch this patch is enabling LoopBack users to write their apps in TypeScript. We may want to prioritize this feature in our backlog.

PR #214, opened 2016-09-22
Update deps to loopback 3.0.0 RC

bajtos:
Connect to strongloop-internal/scrum-loopback#1063

cc @davidcheung I think we should create a `loopback-2.x` branch in this repo as we are already doing in other projects. It can wait until your big refactor is landed though.


bajtos:
@davidcheung the tests are failing because they use a feature that's no longer available in LoopBack 3.0. Do you have bandwidth to take a look?

PR #216, opened 2016-09-23
Normalize line endings to support both LF and CRLF [2.x]

bajtos:
This should fix build failures on Windows caused by line-ending mismatch.

Partial back-port of #214

cc @davidcheung


bajtos:
Closed in favour of #217.

PR #222, opened 2016-10-12
Fix resolution of middleware modules [2.x]

bajtos:
Fix the code loading "loopback#errorhandler" (for example) to correctly look up the "loopback" module in node_modules of the bootstrapped application, instead of looking it up in node_modules of loopback-boot.

Note that I am intentionally not including any unit-tests, because setting up a correct directory layout reproducing the problem is too difficult.

@davidcheung please review.

I am intentionally sending the PR against the 2.x branch, please forward-port it to your refactor pull request #181.

cc @kraman This should fix the issue you reported in our chat.
cc @raymondfeng

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -569,7 +569,7 @@ function resolveMiddlewarePath(rootDir, middleware, config) {
 
   // Try to require the module and check if <module>.<fragment> is a valid
   // function
-  var m = require(pathName);
+  var m = require(sourceFile);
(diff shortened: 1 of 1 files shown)


davidcheung:
@slnode test please

kraman:
👍  Tested against my installation and fix works. Thanks @bajtos

bajtos:
Released in `loopback-boot@2.23.0`, enjoy :)

PR #239, opened 2017-02-14
[Feature] Register the new config file formats for the model config file

snowyu:
You can use your favorite config file formats to write model, by register these formats first:

server.js

```
var loadConfig = require('load-config-file');
var yaml  = require('js-yaml');
var cson  = require('cson');

loadConfig.register(['.yaml', '.yml'], yaml.safeLoad); //first search.
loadConfig.register('.cson', cson.parseCSONString.bind(cson)); //second search


var loopback = require('loopback');
var boot = require('loopback-boot');

var app = module.exports = loopback();

//....

```

Note: It CANNOT load script for such config file if you have registered the config format to `require.extensions`.
See /lib/utils.js#118


slnode:

slnode:

slnode:

slnode:

anodynos:
It works with `common/models/someModel.cson` but NOT with other configs apparently.
I tried with `server/component-config.cson` and it didn't pick it up :-(

I'ts a great feature, please fix and merge!

anodynos:
Also the generators of the cli dont work any more - i.e `lb relation` doesn't recognise the .cson relations :-(

snowyu:
Only for model config currently.

slnode:

slnode:

snowyu:
Now the unit test `compiler from directory supports sources relative to node_modules` failed on **2.x branch**.

```
  206 passing (15s)
  1 failing

  1) compiler from directory supports sources relative to node_modules:

      AssertionError: expected { Object (name, config, ...) } to deeply equal { Object (name, config, ...) }
      + expected - actual

               "required": true
               "type": "string"
             }
             "emailVerified": "boolean"
      +      "id": {
      +        "generated": true
      +        "id": 1
      +        "type": [Function]
      +      }
             "lastUpdated": "date"
             "password": {
               "required": true
               "type": "string"
```

Reason:

The `loopback-boot` use the cached object(require .json) to process, so the unit test could require the same json object(only load from file once).  And the `load-config-file` has always loaded file to create a new json object. 

But I wonder why this happen? should it somewhere else use the `require` to load model config again?

snowyu:
@anodynos now add the config file formats supports to any configs on plugin-base for **master branch**

snowyu:
**2.x branch** done to support all configs.

PR #240, opened 2017-03-08
Provide scriptExtensions option for 3.x

supasate:
### Description
Provide `scriptExtensions` option for custom loading.
Due to `require.extensions` is not used by some framework (e.g. `jest`). So, this feature allows specifying file extensions supported by custom loaders.

#### Related issues
strongloop/loopback#3204

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #242, opened 2017-03-23
Provide options.scriptExtensions for 2.x

supasate:
### Description
Provide scriptExtensions option for custom loading.
Due to require.extensions is not used by some framework (e.g. jest). So, this feature allows specifying file extensions supported by custom loaders.

#### Related issues
strongloop/loopback#3204

### Checklist

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -17,6 +17,13 @@ var g = require('strong-globalize')();
 
 var FILE_EXTENSION_JSON = '.json';
 
+function arrayToObject(array) {
+  return array.reduce(function(obj, val) {
+    obj[val] = val;
+    return obj;
+  }, {});
+}
+
 /**
  * Gather all bootstrap-related configuration data and compile it into
  * a single object containing instruction for `boot.execute`.
@@ -36,8 +43,14 @@ module.exports = function compile(options) {
     options = { appRootDir: options };
   }
 
+  // For setting properties without modifying the original object
+  options = Object.create(options);
+
   var appRootDir = options.appRootDir = options.appRootDir || process.cwd();
   var env = options.env || process.env.NODE_ENV || 'development';
+  var scriptExtensions = options.scriptExtensions ?
+    arrayToObject(options.scriptExtensions) :
+    require.extensions;
 
   var appConfigRootDir = options.appConfigRootDir || appRootDir;
   var appConfig = options.config ||
@@ -76,9 +89,9 @@ module.exports = function compile(options) {
   resolveRelativePaths(bootScripts, appRootDir);
 
   bootDirs.forEach(function(dir) {
-    bootScripts = bootScripts.concat(findScripts(dir));
+    bootScripts = bootScripts.concat(findScripts(dir, scriptExtensions));
     var envdir = dir + '/' + env;
-    bootScripts = bootScripts.concat(findScripts(envdir));
+    bootScripts = bootScripts.concat(findScripts(envdir, scriptExtensions));
   });
 
   // de-dedup boot scripts -ERS
@@ -90,12 +103,12 @@ module.exports = function compile(options) {
 
   var modelSources = options.modelSources || modelsMeta.sources || ['./models'];
   var modelInstructions = buildAllModelInstructions(
-    modelsRootDir, modelsConfig, modelSources, options.modelDefinitions);
+    modelsRootDir, modelsConfig, modelSources, options.modelDefinitions,
+    scriptExtensions);
 
-  var mixinDirs = options.mixinDirs || [];
   var mixinSources = options.mixinSources || modelsMeta.mixins || ['./mixins'];
   var mixinInstructions = buildAllMixinInstructions(
-    appRootDir, mixinDirs, mixinSources, options, modelInstructions);
+    appRootDir, options, mixinSources, scriptExtensions, modelInstructions);
 
   // When executor passes the instruction to loopback methods,
   // loopback modifies the data. Since we are loading the data using `require`,
@@ -151,11 +164,11 @@ function assertIsValidModelConfig(config) {
  * @private
  */
 
-function findScripts(dir, extensions) {
+function findScripts(dir, scriptExtensions) {
   assert(dir, g.f('cannot require directory contents without directory name'));
 
   var files = tryReadDir(dir);
-  extensions = extensions || _.keys(require.extensions);
+  scriptExtensions = scriptExtensions || require.extensions;
 
   // sort files in lowercase alpha for linux
   files.sort(function(a, b) {
@@ -181,9 +194,9 @@ function findScripts(dir, extensions) {
     var filepath = path.resolve(path.join(dir, filename));
     var stats = fs.statSync(filepath);
 
-    // only require files supported by require.extensions (.txt .md etc.)
+    // only require files supported by specified extensions
     if (stats.isFile()) {
-      if (isPreferredExtension(filename))
+      if (scriptExtensions && isPreferredExtension(filename, scriptExtensions))
         results.push(filepath);
       else
         debug('Skipping file %s - unknown extension', filepath);
@@ -204,9 +217,12 @@ function tryReadDir() {
 }
 
 function buildAllModelInstructions(rootDir, modelsConfig, sources,
-    modelDefinitions) {
-  var registry = verifyModelDefinitions(rootDir, modelDefinitions) ||
-    findModelDefinitions(rootDir, sources);
+                                   modelDefinitions, scriptExtensions) {
+  var registry = verifyModelDefinitions(rootDir, modelDefinitions,
+                                        scriptExtensions);
+  if (!registry) {
+    registry = findModelDefinitions(rootDir, sources, scriptExtensions);
+  }
 
   var modelNamesToBuild = addAllBaseModels(registry, Object.keys(modelsConfig));
 
@@ -287,7 +303,7 @@ function sortByInheritance(instructions) {
     });
 }
 
-function verifyModelDefinitions(rootDir, modelDefinitions) {
+function verifyModelDefinitions(rootDir, modelDefinitions, scriptExtensions) {
   if (!modelDefinitions || modelDefinitions.length < 1) {
     return undefined;
   }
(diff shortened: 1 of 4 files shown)


slnode:

slnode:

slnode:

slnode:

bajtos:
@supasate thank you for the pull request! Does @clarkorz's https://github.com/strongloop/loopback-boot/pull/240#discussion_r107839885 apply here too?

> The third param `onlyScriptsExportingFunction` cause `.js` files of models will not be loaded. 
> Line 142 in utils.js checking `typeof require.extensions['.js'] === 'function'` which will be false when using jest.

bajtos:
@slnode  ok to test

lib/compiler.js
```diff
@@ -38,6 +38,8 @@ module.exports = function compile(options) {
 
   var appRootDir = options.appRootDir = options.appRootDir || process.cwd();
   var env = options.env || process.env.NODE_ENV || 'development';
+  var scriptExtensions = options.scriptExtensions = options.scriptExtensions ||
```
bajtos: Modifying input arguments (setting `options.scriptExtensions`) is a bad practice, we have been bitten by this in the past. Please treat `options` are readonly/frozen.

I see that you are accessing `options.scriptExtensions` at https://github.com/strongloop/loopback-boot/pull/242/files#diff-ee97c5091b89979aace94674818996baR722.

Here are two of many possible solutions how to keep `options` frozen and still have `scriptExtensions` set:

```js
// use prototypal inheritance
options = Object.create(options);

// shallow-copy properties
options = util._extend({}, options);
```

(A line like the above would go to L38-39 above, before we start processing the `options` object.)

lib/compiler.js
```diff
@@ -204,9 +209,10 @@ function tryReadDir() {
 }
 
 function buildAllModelInstructions(rootDir, modelsConfig, sources,
-    modelDefinitions) {
-  var registry = verifyModelDefinitions(rootDir, modelDefinitions) ||
-    findModelDefinitions(rootDir, sources);
+    modelDefinitions, scriptExtensions) {
+  var registry = verifyModelDefinitions(rootDir, modelDefinitions,
+                  scriptExtensions) ||
+                  findModelDefinitions(rootDir, sources, scriptExtensions);
```
bajtos: The indentation makes it difficult to distinguish function parameters from the second function call. I am proposing the following style:

```js
  var registry = verifyModelDefinitions(rootDir, modelDefinitions,
                                        scriptExtensions);

  if (!registry) {
    registry = findModelDefinitions(rootDir, sources, scriptExtensions);
  }
```

lib/compiler.js
```diff
@@ -38,6 +38,8 @@ module.exports = function compile(options) {
 
   var appRootDir = options.appRootDir = options.appRootDir || process.cwd();
   var env = options.env || process.env.NODE_ENV || 'development';
+  var scriptExtensions = options.scriptExtensions = options.scriptExtensions ||
+                                                    _.keys(require.extensions);
```
bajtos: Considering that you are converting this  array back into an object down in `isPreferredExtension`, I think it's better to make the conversion right here.

```js
var scriptExtensions = options.scriptExtensions ?
  arrayToObject(scriptExtensions) : 
  require.extensions;
```

lib/compiler.js
```diff
@@ -644,19 +657,28 @@ function getExcludedExtensions() {
...
+function arrayToObject(array) {
+  return array.reduce(function(obj, val) {
+    obj[val] = val;
+    return obj;
+  }, {});
+}
+
+function isPreferredExtension(filename, extensions) {
+  var includeExtensions = extensions ?
+    arrayToObject(extensions) : require.extensions;
```
bajtos: Since `isPreferredExtension` is a private method, there is no need to keep support for the old one-arg API.

```js
function isPreferredExtension(filename, includeExtensions) {
  assert(!!includeExtensions, '"includeExtensions" argument is required');
  // the code below remains unchanged
  var ext = path.extname(filename);
  return (ext in includeExtensions) && !(ext in getExcludedExtensions());}
```

test/fixtures/simple-app/boot/custom.customjs
```diff
@@ -0,0 +1,8 @@
+process.bootFlags.push('customLoaded');
+module.exports = function(app, callback) {
+  process.bootFlags.push('customStarted');
+  process.nextTick(function() {
+    process.bootFlags.push('customFinished');
+    callback();
+  });
```
bajtos: This is unnecessarily complex. The thing to verify in your test is whether loopback-boot recognises a custom extension. There is no need to verify how boot script are loaded/executed again.

```js
'use strict';

module.exports = function(app, callback) {
  process.bootFlags.push('customjs');
  callback();
};
```

Same comment applies to `custom.customjs2`.

test/executor.test.js
```diff
@@ -299,6 +299,27 @@ describe('executor', function() {
       });
     });
 
+    it('should load scripts specified in options.scriptExtensions',
```
```js
it('searches boot file extensions specified in options.scriptExtensions')
```

See also http://loopback.io/doc/en/contrib/style-guide.html#test-naming

bajtos:
@supasate also please take a look at https://github.com/strongloop/loopback-boot/pull/240#discussion_r107839885 - please add a test reproducing the issue (I think the trick will be to use a model file with a custom extension that's not registered in `required.extensions`.)

bajtos:
Also please rebase your patch on top of the latest `2.x` branch, I fixed the failing Node 0.10 builds in #244.

supasate:
Hi @bajtos, I've rebased and fixed issues based on your comments. I made a sample repo at https://github.com/supasate/loopback-boot-with-jest. I'll squash later if it's alright.

supasate:
I also found another issue about file name resolving (https://github.com/strongloop/loopback/issues/2201) (not related to this PR but it will make `jest` failed to run).

lib/compiler.js
```diff
@@ -670,51 +699,52 @@ function fixFileExtension(filepath, files, onlyScriptsExportingFunction) {
...
+          //results.push(otherFile);
+        //}
       }
     }
   });
   return (results.length > 0 ? results[0] : undefined);
 }
 
-function resolveAppScriptPath(rootDir, relativePath, resolveOptions) {
+function resolveAppScriptPath(rootDir, relativePath, resolveOptions,
```
bajtos: Please keep `resolveOptions` as the last function arg. I would consider converting `scriptExtensions` into a property in `resolveOptions`, to keep the number of function arguments reasonably short.

lib/compiler.js
```diff
@@ -670,51 +699,52 @@ function fixFileExtension(filepath, files, onlyScriptsExportingFunction) {
...
-function buildAllMixinInstructions(appRootDir, mixinDirs, mixinSources, options,
-  modelInstructions) {
-  var extensions = _.without(_.keys(require.extensions),
-    _.keys(getExcludedExtensions()));
-
+function buildAllMixinInstructions(appRootDir, mixins, mixinDirs, mixinSources,
+                                   scriptExtensions, normalization,
+                                   modelInstructions) {
   // load mixins from `options.mixins`
-  var sourceFiles = options.mixins || [];
```
bajtos: What is the rationale for this change? I mean why to pass `options` properties as individual arguments? `buildAllMixinInstructions` is already accepting way too many arguments, we should look into ways how to make that list shorter, not longer...

lib/compiler.js
```diff
@@ -670,51 +699,52 @@ function fixFileExtension(filepath, files, onlyScriptsExportingFunction) {
 
       if (!(otherFileExtension in getExcludedExtensions()) &&
         path.basename(f, otherFileExtension) == basename) {
-        if (!onlyScriptsExportingFunction)
-          results.push(otherFile);
-        else if (onlyScriptsExportingFunction &&
-          (typeof require.extensions[otherFileExtension]) === 'function') {
-          results.push(otherFile);
-        }
+        // if (!onlyScriptsExportingFunction)
```
bajtos: We don't keep commented-out code around, please remove.

If I understand this code correctly, the intention was:

 - return a file with any extension when `onlyScriptsExportingFunction` is not set
 - return a file with an extension that can be loaded by Node.js runtime (an extension that is registered in `require.extensions`) when `onlyScriptsExportingFunction` is set

We should preserve these two modes.

I am proposing to replace `onlyScriptsExportingFunction` arg with `scriptExtensions` and modify all function callers accordingly.

```js
// implementation here
if (!scriptExtensions || otherFileExtension in scriptExtensions)
  results.push(otherFile);

// usage - flag not set
fixFileExtension(filepath, files)

// usage -flag set
fixFileExtension(filepath, files, scriptExtensions)
```

bajtos:
@supasate thank you for the pull request, please take a look at the comment I left above. Also please rebase your patch on top of the latest `2.x` branch.

lib/compiler.js
supasate: Oops! That's my mistake. I tested something and forgot to uncomment. I'll fix as you suggested.

lib/compiler.js
supasate: I tried to make code look consistent with [how `mixinDirs` and `mixinSources` are declared  and passed to the function.](https://github.com/supasate/loopback-boot/blob/8fef5f6fa8b0d2e37b1c38877805fbfeb656dbd6/lib/compiler.js#L112-L113).

Do you want to refactor those arguments by passing only `options` and set their default values in `buildAllMixinInstructions` instead?

supasate:
About rebasing on top of the latest `2.x` branch, do you mean latest of `2.x` branch or `2.x-latest` branch? (I've rebased with `2.x` branch because it seems its last commit is more up-to-date than `2.x-latest`.)

lib/compiler.js
supasate: After refactoring, there is no need to use `scriptExtensions` in this function anymore.

lib/compiler.js
supasate: I refactored by passing only `options`.

bajtos:

bajtos:
Released in `loopback-boot@2.24.0` 🎉

supasate:
Thanks @bajtos ! 

I think you can review the 3.x branch also. The code is almost the same except that it is splitted into several files.

PR #252, opened 2017-07-24
Add support for ES6 style async boot scripts (2.x)

lehni:
### Description

This PR allows boot scripts to be defined in the ES6 async fashion, without the need for passing callbacks:

```js
module.exports = async function(app) {
   ...
};
```

#### Related issues

- None

### Checklist

- [ ] New tests added or existing tests modified to cover all changes

### DIFF
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -305,17 +305,24 @@ function runScripts(app, list, callback) {
 
   async.eachSeries(functions, function(f, done) {
     debug('Running script %s', f.path);
-    if (f.func.length >= 2) {
-      debug('Starting async function %s', f.path);
-      f.func(app, function(err) {
-        debug('Async function finished %s', f.path);
-        done(err);
-      });
-    } else {
-      debug('Starting sync function %s', f.path);
-      f.func(app);
-      debug('Sync function finished %s', f.path);
-      done();
+    var cb = function(err) {
+      debug('Async function finished %s', f.path);
+      done(err);
+      // Make sure done() isn't called twice, e.g. if a script returns a
+      // thenable object and also calls the passed callback.
+      cb = null;
+    };
+    try {
+      var result = f.func(app, cb);
+      if (result && typeof result.then === 'function') {
+        result.then(function() { cb(); }, cb);
+      } else if (f.func.length < 2) {
+        debug('Sync function finished %s', f.path);
+        done();
+      }
+    } catch (err) {
+      debug('Sync function failed %s', f.path, err);
+      done(err);
     }
   }, callback);
 }
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -265,6 +265,8 @@ describe('executor', function() {
         'barLoaded',
         'barSyncLoaded',
         'fooLoaded',
+        'promiseLoaded',
+        'thenableLoaded',
         'barStarted',
       ]);
 
@@ -275,9 +277,15 @@ describe('executor', function() {
           'barLoaded',
           'barSyncLoaded',
           'fooLoaded',
+          'promiseLoaded',
+          'thenableLoaded',
           'barStarted',
           'barFinished',
           'barSyncExecuted',
+          'promiseStarted',
+          'promiseFinished',
+          'thenableStarted',
+          'thenableFinished',
           'umdLoaded',
         ]);
         done();
@@ -292,9 +300,15 @@ describe('executor', function() {
           'barLoaded',
           'barSyncLoaded',
           'fooLoaded',
+          'promiseLoaded',
+          'thenableLoaded',
           'barStarted',
           'barFinished',
           'barSyncExecuted',
+          'promiseStarted',
+          'promiseFinished',
+          'thenableStarted',
+          'thenableFinished',
           'umdLoaded',
         ]);
         done();
@@ -305,7 +319,7 @@ describe('executor', function() {
     function(done) {
       var options = {
         app: app,
-        appRootDir: path.join(__dirname, './fixtures/simple-app'),
+        appRootDir: SIMPLE_APP,
         scriptExtensions: ['.customjs', '.customjs2'],
       };
       boot.execute(app, boot.compile(options), function(err) {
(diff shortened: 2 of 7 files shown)


slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

lib/executor.js
```diff
@@ -305,12 +306,18 @@ function runScripts(app, list, callback) {
...
+    if (f.func.length >= 2 || isAsync) {
       debug('Starting async function %s', f.path);
-      f.func(app, function(err) {
+      var cb = function(err) {
         debug('Async function finished %s', f.path);
         done(err);
-      });
+      };
+      if (isAsync) {
+        f.func(app).then(cb.bind(null, null)).catch(cb);
```
bajtos: If `cb(null)` throws an error, then `.catch()` will invoke `cb` again with this error - that's not correct IMO. You should pass the error-handling-callback as the second argument of `.then()`:

```js
f.func(app).then(cb.bind(null, null), cb);
```

bajtos:
This is a great feature, I am looking forward to get it added! 👍

lehni:
@bajtos I will work on this improvements and unit test for them. In the meantime, can you tell me which branch I should work on? Since there are two active branches (2.x and master), and I am still using 2.x for LB 3 (master doesn't seem to load mixins correctly, I meant to create an issue for this), I guess I'll work on 2.x first and then port to master when it landed?

bajtos:
> I guess I'll work on 2.x first and then port to master when it landed?

Yes please, this works best for me.

lehni:
@bajtos I've pushed a new version based on your input, along with some unit tests. PTAL

lehni:
Two more things:

1. I needed to use an anonymous function instead of the `=>`, because eslint was complaining about it. I tried to update eslint to a version that supports it, but that triggered a whole row of other linting issues, so I stuck with the function for now

2. The tests reference the `Promise` object. I am not sure if that's fine? If not, then I can keep it all to `thenable` objects.

test/fixtures/simple-app/boot/promise.reject
```diff
@@ -0,0 +1,8 @@
+// Copyright IBM Corp. 2017. All Rights Reserved.
+// Node module: loopback-boot
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+module.exports = function(app) {
+  return Promise.reject('Promise.reject()')
```
bajtos: Our practice is to reject Promises with Error instances.

```js
return Promise.reject(new Error('Promise.reject()'));
```

bajtos: Also please add `.js` extension to the filename. I am surprised your new tests are passing despite this problem?

lib/executor.js
```diff
@@ -305,15 +305,20 @@ function runScripts(app, list, callback) {
...
+    var cb = function(err) {
+      debug('Async function finished %s', f.path);
+      done(err);
+    };
+    var result, error;
+    try {
+      result = f.func(app, cb);
+    } catch (err) {
+      debug('Sync function failed %s', f.path, err);
+      error = err;
```
bajtos: AFAICT, `error` is not used. Can we remove it please?

Also when the function exported by a boot script fails, the error should be reported back/up to the caller of `boot(app)`. I think you should call `return done(err)` here.

bajtos:
@lehni the CI is failing on Node.js 0.10, see https://travis-ci.org/strongloop/loopback-boot/jobs/258394078. It may be caused by the fact that Node.js 0.10 does not have a built-in `Promise` object, in which case you can add `bluebird` as a dev-dependency and use bluebird Promise instead of the built-in one in the newly added boot scripts. Also because your new code is silently swallowing errors thrown by boot scripts, the CI output is not making the problem obvious.

lehni:
@bajtos oops, yeah that's why the errors didn't make sense to me probably. I will look into both issues right now. Stil learning :)

test/fixtures/simple-app/boot/promise.reject
lehni: I am using a bit of a workaround with `scriptExtensions` set to `'.reject'` for this test, so the failing boot script does not get loaded in all other tests, but I can then explicitly request it for that single test and produce the rejection. It's not super clean but I couldn't think of another way to achieve this goal without creating another full project... I hope it's fine?

lib/executor.js
lehni: It was late last night... My bad. Will address it now.

test/fixtures/simple-app/boot/promise.reject
lehni: Actually, I think I have a better idea for how to control this. I'll do the same to test for sync exceptions also.

lehni:
@bajtos I think this is it. I responded to all your feedback and added a test for errors thrown from boot scripts. And it looks like it's passing all tests!

lib/executor.js
```diff
@@ -305,17 +305,24 @@ function runScripts(app, list, callback) {
...
+      debug('Async function finished %s', f.path);
+      done(err);
+      // Make sure done() isn't called twice, e.g. if a script returns a
+      // thenable object and also calls the passed callback.
+      cb = null;
+    };
+    try {
+      var result = f.func(app, cb);
+      if (result && typeof result.then === 'function') {
+        result.then(function() { cb && cb(); }, cb);
```
bajtos: `cb` is always defined, the check `cb &&` is redundant.

test/executor.test.js
```diff
@@ -317,44 +331,84 @@ describe('executor', function() {
...
-          'module.exports = ' +
-          'function(Model, options) {}');
-
-        appdir.writeFileSync('custom-mixins/time-stamps.js',
-          'module.exports = ' +
-          'function(Model, options) {}');
+  describe('with boot script returning a rejected promise', function() {
+    before(function() {
+      // Tell simple-app/boot/reject.js to return a rejected promise
+      process.rejectPromise = true;
```
bajtos: It would be probably simpler to create a different application (a set of instructions) for each of these error cases. Having said that, your approach with process flags is ok too.

bajtos:

bajtos:
Released as `loopback-boot@2.26.0`. Enjoy 🎉

lehni:
Very cool! That went faster than I'd imagined : )

lehni:
Shall I port this to #253 now?

slnode:

lehni:
@bajtos the reason for `cb && cb()` was in this comment above:
```js
    var cb = function(err) {
      done(err);
      // Make sure done() isn't called twice, e.g. if a script returns a
      // thenable object and also calls the passed callback.
      cb = null;
    };
```
I believe it's still valid?

bajtos:
> the reason for cb && cb() was in this comment above:
> (...)
> I believe it's still valid?

@lehni Ouch! You are right. I think a better solution is to set `cb` to a no-op function instead of `null`. Would you mind contributing this change in a new pull request? It would be great to add a unit-test with a script returning both a thenable object and calling the callback.

> Shall I port this to #253 now?

Yes please 👍

lehni:
@bajtos I will wait for #260 before resolving #253, and will then roll both #252 and #260 into it.

PR #260, opened 2017-08-15
Do not call callbacks twice in async boot scripts

lehni:
### Description

As discussed in https://github.com/strongloop/loopback-boot/pull/252#issuecomment-319906102, we need to make sure that a boot script that both returns a promise and calls the callback method does not trigger two calls of the callback method passed to `boot.execute()`.

#### Related issues

- #252

### Checklist

### DIFF
--- a/test/executor.test.js
+++ b/test/executor.test.js
@@ -373,6 +373,23 @@ describe('executor', function() {
     });
   });
 
+  describe('with boot script returning a promise and calling callback',
+    function() {
+      before(function() {
+        process.promiseAndCallback = true;
+      });
+
+      after(function() {
+        delete process.promiseAndCallback;
+      });
+
+      it('should only call the callback once', function(done) {
+        // Note: Mocha will fail this test if done() is called twice
+        boot.execute(app, simpleAppInstructions(), done);
+      });
+    }
+  );
+
   describe('for mixins', function() {
     var options;
     beforeEach(function() {
--- a/lib/executor.js
+++ b/lib/executor.js
@@ -306,11 +306,11 @@ function runScripts(app, list, callback) {
   async.eachSeries(functions, function(f, done) {
     debug('Running script %s', f.path);
     var cb = function(err) {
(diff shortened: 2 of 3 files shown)


slnode:

slnode:

slnode:

slnode:

test/executor.test.js
```diff
@@ -373,6 +373,31 @@ describe('executor', function() {
...
+
+      it('should only call the callback once',
+      function(done) {
+        var count = 0;
+        boot.execute(app, simpleAppInstructions(), function(err) {
+          if (err) return done(err);
+          count++;
+          expect(count).to.eql(1);
+          process.nextTick(function() {
+            done();
```
bajtos: I think this is rather brittle as it relies on `boot.execute` to make the second call of the callback before `process.nextTick` is run.

Fortunately, mocha detects the situation when the `done` callback was called twice and fails the test with a helpful error: `Error: done() called multiple times`.

I am proposing to simplify this test to simply pass `done` as the callback of `boot.execute`.

bajtos:
@slnode ok to test

test/executor.test.js
lehni: Ok, let me test this.

lehni: You're right! I adjusted the PR accordingly now.

test/executor.test.js
```diff
@@ -373,6 +373,22 @@ describe('executor', function() {
...
+      before(function() {
+        process.promiseAndCallback = true;
+      });
+
+      after(function() {
+        delete process.promiseAndCallback;
+      });
+
+      it('should only call the callback once', function(done) {
+        boot.execute(app, simpleAppInstructions(), done);
```
bajtos: Perhaps add a comment explaining that mocha fails the test when the callback was called multiple times?

bajtos:
Landed, thank you for the contribution! 🎉

PR #256, opened 2017-07-26
[BUG] Middleware is loading .js.map of .js

ivanproskuryakov:
https://github.com/strongloop/loopback-boot/issues/255

Issue #255, opened 2017-07-26
[BUG] Middleware is loading .js.map of .js


ivanproskuryakov:
The Problem comes when a custom middleware based on Typescript is used. Typescript generates two files `.js` and `.js.map` - which is used for mapping.
![errorhandler js map - resport today - www resport dev resport today 2017-07-26 12-08-38](https://user-images.githubusercontent.com/3517364/28613531-43b3bd5e-71fb-11e7-8983-176928eee87e.png)

`map.js` should be also covered in `getExcludedExtensions` method https://github.com/strongloop/loopback-boot/blob/master/lib/utils.js#L101

Otherwise, its breaking the execution with the error:
```sh
Unhandled rejection SyntaxError: Unexpected token :
/Users/ivan/www/resport.dev/resport.today/build/server/middleware/errorLogger.js.map:1
(function (exports, require, module, __filename, __dirname) { {"version":3,"sources":["server/middleware/errorLogger.ts"],"names":[],"mappings":";AAAA,6CAA0C;AAK1C,iBAAS,cAAM,OAAA,UAAC,GAAG,EAAE,GAAG,EAAE,GAAG,EAAE,IAAI;IACjC,IAAI,eAAM,CAAC,GAAG,CAAC,GAAG,CAAC,CAAC,KAAK,CAAC,GAAG,CAAC,CAAC;IAC/B,MAAM,CAAC,IAAI,CAAC,GAAG,CAAC,CAAC;AACnB,CAAC,EAHc,CAGd,CAAC","file":"errorLogger.js","sourcesContent":["import {Logger} from 'app/service/logger';\n\n/**\n * error logger middleware\n */\nexport = () => (err, req, res, next) => {\n  new Logger(req.app).error(err);\n  return next(err);\n};\n"]}
                                                                        ^
SyntaxError: Unexpected token :
    at Object.exports.runInThisContext (vm.js:76:16)
    at Module._compile (module.js:513:28)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Module.require (module.js:468:17)
    at require (internal/module.js:20:19)
    at /Users/ivan/www/resport.dev/resport.today/node_modules/loopback-boot/lib/plugins/middleware.js:259:19
    at Array.forEach (native)
[...]

PR #245, opened 2017-04-11
Exclude .map extension

dtomasi:
### Description
If you are using loopback with typescript or any other javascript dialect, it would generate the .map files for the debugger. loopback-boot tries to load these files which ends up into a Syntax-Error.

#### Related issues

- #246

### Checklist

- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #263, opened 2017-09-14
Ignore js sourcemap files from boot

zbarbuto:
### Description

See discussion from #256  and #245

#### Related issues

- #255

### Checklist

cc @bajtos

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -661,6 +661,12 @@ function resolveRelativePaths(relativePaths, appRootDir) {
 function getExcludedExtensions() {
   return {
     '.json': '.json',
+    /**
+     * This is a temporary workaround for #246
+     * See discussion here for full description of the underlying issue
+     * https://github.com/strongloop/loopback-boot/pull/245#issuecomment-311052798
+     */
+    '.map': '.map',
     '.node': 'node',
   };
 }
--- a/test/compiler.test.js
+++ b/test/compiler.test.js
@@ -2187,6 +2187,24 @@ describe('compiler', function() {
         'sourceFile', coffee);
     });
 
+    it('ignores sourcmap files when loading middleware',
+    function() {
+      var middleware = appdir.writeFileSync('my-middleware.js',
+        '// I am the middleware');
+      var sourcemap = appdir.writeFileSync('my-middleware.js.map',
+        '// I am a sourcemap');
+      appdir.writeConfigFileSync('middleware.json', {
+        'routes': {
+          './my-middleware': {},
+        },
+      });
+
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

lehni:
I haven't been part of the previous discussions in #256, but this does look good to me. Let's see what @bajtos thinks, since he pointed out some problems in that issue.

@zbarbuto don't forget to sign the [CLA](https://cla.strongloop.com/agreements/strongloop/loopback-boot), as otherwise the checks won't pass and we won't be able to merge.

zbarbuto:
Whoops. Forget I need to sign it for each individual SL repo I contribute to.

bajtos:
@slnode ok to test

lib/compiler.js
```diff
@@ -661,6 +661,7 @@ function resolveRelativePaths(relativePaths, appRootDir) {
 function getExcludedExtensions() {
   return {
     '.json': '.json',
+    '.map': '.map',
```
bajtos: Please add a comment explaining that this is a temporary workaround - see https://github.com/strongloop/loopback-boot/pull/245#issuecomment-329098622

> I think the `.map` entry should be also accompanied by a comment explaining that it's a partial fix for a bigger problem and pointing to https://github.com/strongloop/loopback-boot/pull/245#issuecomment-311052798

test/compiler.test.js
```diff
@@ -2187,6 +2187,26 @@ describe('compiler', function() {
...
+        'routes': {
+          './my-middleware': {},
+        },
+      });
+
+      var instructions = boot.compile(appdir.PATH);
+
+      expect(instructions.middleware.middleware[0]).have.property(
+        'sourceFile', middleware);
+      expect(instructions.middleware.middleware[0]).not.have.property(
```
bajtos: This second expectation is redundant - if the `sourceFile` property is set to `middleware`, then the second expectation will always pass. If `sourceFile` is not set to `middleware`, then the first `expect` will abort the test and this second expectation won't be run at all.

Please remove these two lines (2206-2207).

zbarbuto: Ah, yes, fair point. My goal was to ensure that the map file is definitely never loaded but I missed the index being used. Will remove.

lib/compiler.js
zbarbuto: Ah, yep, missed that. Cheers.

bajtos:
Released in `loopback-boot@2.26.2`, thank you for the contribution!

PR #269, opened 2017-10-16
update strong-globalize to 3.1.0

shimks:
Testing whether the 2.x branch works with the latest version of strong-globalize

- connect to update strong-globalize to 3.1.0 #268

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "debug": "^2.0.0",
     "lodash": "^3.6.0",
     "semver": "^4.1.0",
-    "strong-globalize": "^2.6.2",
+    "strong-globalize": "^3.1.0",
     "toposort": "^0.2.10"
   },
   "devDependencies": {


dhmlau:
@bajtos, looks like we have errors in running node 0.10 and 0.12.  I'd like to propose we update the `debug` version in strong-globalize in 2.x and make the 2.x loopback-* module using that version instead of `strong-globalize` 3.x.  What do you think?

```
/home/jenkins/workspace/nb/loopback-boot~2.x/4b9e0b16/node_modules/strong-globalize/node_modules/os-locale/index.js:2
const execa = require('execa');
^^^^^
SyntaxError: Use of const in strict mode.
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
```

bajtos:
Yeah, we will need to drop support for Node.js 0.x first, before this patch can be landed. See what I started in strong-remoting: https://github.com/strongloop/strong-remoting/pull/428

Alternatively, and I think this is actually a better option, we can fix strong-globalize's 2.x version line to ignore the Snyk vulnerability that's not affecting us and then publish a new version of strong-globalize. It may result in less work overall, because all dependents of strong-globalize@2 will start passing snyk/nsp checks with no modifications needed.

Thoughts?

dhmlau:
sounds good to me.  
@shimks, in this case, we'd need to add `.nsprc` file in strong-globalize 2.x, same as what we did in master.
https://github.com/strongloop/strong-globalize/blob/master/.nsprc

bajtos:
I opened #272 to remove support for Node.js 0.10 and 0.12 in our 2.x branch.

shimks:
@slnode test please

shimks:
@slnode test please

dhmlau:
@shimks @bajtos , if this change is only needed for getting rid of the security vulnerability, this is no longer needed.

PR #282, opened 2018-07-23
chore: update dependencies

dhmlau:
### Description
Update dependencies in package.json

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: node_js
 node_js:
-  - "4"
   - "6"
-  - "7"
+  - "8"
+  - "10"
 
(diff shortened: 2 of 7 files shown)


jannyHou:
@dhmlau The datasources of the 4 tests in [dynamically configure the datasource](https://github.com/strongloop/loopback-boot/blob/master/test/executor.test.js#L1182) are missing `{connector: 'connector_name'}`, a quick fix would be specify the `memory` connector for them.

E.g.:
Change [the datasource here](https://github.com/strongloop/loopback-boot/blob/master/test/executor.test.js#L1189-L1194) to be 
```js
var datasource = {
        mydb: {
          connector: 'memory',
          host: '${DYNAMIC_HOST}',
          port: '${DYNAMIC_PORT}',
        },
};
```

It takes time to create a fake connector, so I would suggest just use the `memory`.

dhmlau:
Thanks @jannyHou!  I just submitted a change to drop Node 4 in Travis.  Will wait for the CI to finish.

PR #287, opened 2019-03-22
Upgrade 2.x deps

raymondfeng:
### Description

Upgrade deps to remove `npm audit` warnings.

#### Related issues

- connect to <link_to_referenced_issue>

### Checklist


test/helpers/browser.js
```diff
@@ -8,7 +8,8 @@ var vm = require('vm');
 function createContext() {
   var context = {
     // required by browserify
-    XMLHttpRequest: function() { throw new Error('not implemented'); },
```
b-admike: Is this change also associated with the upgrade?

.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=true
```
b-admike: Do we need to enable package-lock here? Is that something we want to do throughout our repos?

test/helpers/browser.js
raymondfeng: Yes. Otherwise, it throws an error - not implemented

.npmrc
raymondfeng: I enabled it to run `npm audit`.

raymondfeng:
FYI: I'm not sure why this PR shows 63 commits.

PR #288, opened 2019-03-22
chore: upgrade deps to fix npm audit complaints

raymondfeng:
#### Related issues

- Recreated from https://github.com/strongloop/loopback-boot/pull/287.

### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - "4"
   - "6"
   - "8"
 
--- a/package.json
+++ b/package.json
@@ -20,29 +20,29 @@
     "lint": "eslint ."
   },
   "engines": {
-    "node": ">=4"
+    "node": ">=6"
   },
   "license": "MIT",
   "dependencies": {
-    "async": "~0.9.0",
-    "commondir": "0.0.1",
-    "debug": "^2.0.0",
-    "lodash": "^4.17.5",
-    "semver": "^4.1.0",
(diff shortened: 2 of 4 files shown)


.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=true
```
b-admike: Following from the other PR, can we remove package-lock file and this change now that we have the upgrades in `package.json`?

raymondfeng: I'm fine either way.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - "4"
```
raymondfeng: Please note it won't work with Node 4.x any more. I find out `loopback@3.x` has `node >= 6`.

package.json
```diff
@@ -20,29 +20,29 @@
...
   "license": "MIT",
   "dependencies": {
-    "async": "~0.9.0",
-    "commondir": "0.0.1",
-    "debug": "^2.0.0",
-    "lodash": "^4.17.5",
-    "semver": "^4.1.0",
-    "strong-globalize": "^2.6.2",
-    "toposort": "^0.2.10"
+    "async": "^2.6.2",
```
bajtos: IIRC, there are subtle but important changes in the way how individual callbacks are scheduled/executed between async versions 0.x and post-1.x.

Unless there is a know security vulnerability in `async@0.9.0`, we should not be upgrading this dependency.

raymondfeng: Reverted.

PR #297, opened 2019-10-03
test: fix browserify-based tests

bajtos:
Add `packageFilter` to handle buggy `async` browserify config.

See the downstream build failure reported for https://github.com/strongloop/loopback/pull/4262 in https://cis-jenkins.swg-devops.com/job/ds/job/loopback-boot%7Emaster/43/console.


bajtos:
I opened a new PR to drop support for Node.js 6.x - see https://github.com/strongloop/loopback-boot/pull/298

